PR #18, opened 2015-04-29
Proper help/usage function

mvdan:


mvdan:
Just switched from flag.Usage to fs.Usage. Now `gb build -h` also shows a proper help :)

davecheney:
Thanks for fixing the usage message but I'm sorry this is not correct. 

The issue is about constructing the usage message from the registered commands, not using a dead blog of text. 

If you want to peal off your other fixes we can merge those seperatly.

mvdan:
You mean having `registerCommand()` affecting the usage string? Would it be ok then to have it also take a short usage string, like "build a package" for "build"?

davecheney:
Yup, you can change whatever you like. We probably need to add

   ShortDesc string
   Desc string

to the Command struct

mvdan:
ShortDesc sure, but Desc? Would that be printed in a new line after ShortDesc?

I was actually thinking of following the `git --help` format, which only contains a sentence (ShortDesc) per command.

davecheney:
If it's of no use at the moment, feel free to omit it.

mvdan:
Updated. I've left a small TODO for the flags specific to commands, since I don't know how to approach that right now.

davecheney:
Nice. Thanks a lot, this is a great improvement.

Issue #37, opened 2015-05-03
gb: tag support via hash


davecheney:
One of the design goals of gb is to handle tags efficiently; any change to the build tags for an `build`/`test` invocation should result in a rebuild of the relevant cached packages.

A subtler twist on this is `$GOOS` and `$GOARCH` are effectively tags, although they are handled more explicitly; eg `$PROJECT/pkg/$GOOS/$GOARCH`. Taking this to it's logical extent the value of `$GOROOT`, and the toolchain used also affect the caching and reuse of cached packages.

I propose that we follow the lead of the `go` tool and use a hash to identify packages constructed from a combination of toolchains, goroots, arches, tags, etc. Unlike the `go` tool, which is embedding this value inside the `.a` file, I propose we put it in a prefix of the package path, as the `isStale` computation uses file paths to assert staleness. Specifically I'm thinking the shorthash of a SHA1 of a set of hashable values.

So, where we currently have, `$PROJECT/pkg/$GOOS/$GOACH/importpath.a`, this would become `$PROJECT/pkg/$HASH/importpath.a`. 

A hash would be computed from a `gb.Context`, and be stable for the use of that `Context`, in theory, the same `Context` constructed

Some (probably not all) the properties of the build that would be:

- $GOROOT 
- toolchain 
- tags
- $GOOS / $GOARCH

The downside of this approach is aggressive use of tags, cross compilation and toolchains would bloat `$PROJECT/pkg` with multiple cached file. I don't see this as a serious concern, this directory is only a cache and can be removed without consequence and rebuilt as needed.

A separate question of how to handle multiple final binaries in `$PROJECT/bin` will have to be addressed in another issue.

bakirov:
> Specifically I'm thinking the shorthash of a SHA1 of a set of hashable values.

What about using non-cryptographic hash functions, like [FNV-1a](http://golang.org/pkg/hash/fnv/)? It seems to be very fast, collisions rare and supported in Go standard library.

davecheney:
Speed isn't a serious concern, the hash is computed from a small number of
items, the path to the compiler, the contents of tag []string, and a few
other things, and computed once per invocation of the build tool.

If it does turn out that creating the build hash is too expensive then we
can fix that later, but I don't think it will be a problem.

mopemope:
Any update on this?

PR #39, opened 2015-05-03
Clarify behavior of -f flag.

jbuberel:
Just a minor change in the documentation for the -f flag, noting that it will disable incremental compilation.


davecheney:
Thanks. Good to be explicit.

davecheney:
It should be noted, that the goal here is to avoid using `-f` to fix problems. I'm hoping the build hash I mentioned in #37 will avoid issues where `$PROJECT/pkg` is stale because it is compiled with another version of the toolchain, via `-goroot`

Issue #46, opened 2015-05-05
main package at $PROJECT/src is ignored


davecheney:
Consider this layout
```
$PROJECT
    src/
        main.go
```
`main.go` will be ignored. Similar to #45, this is a bug, but probably not helpful to the user as the command would be installed as `$PROJECT/bin/src`

davecheney:
As part of discussing #48 it was decided that source code at the root of `$PROJECT/src` was a mistake and the small convenience would only confuse when someone went to consume that package. 

To close this issue, gb should be taught to reject source at the top level of `$PROJECT/src/`

cc @wkennedy

ajpen:
So, what this is saying is any code in $PROJECT/src will be ignored? I have main.go there as I would in a typical go project and its not compiled. My other packages are compiled. Do I need main in a different directory?

davecheney:
Any package at $PROJECT/src is ignored because it would have an invalid package import path.

davecheney:
https://getgb.io/faq/#cannot-build-src-root

PR #48, opened 2015-05-06
Adding support to build straight from 'src'

mariokostelac:
I needed building straight from `src` (#46) so modified code a bit to allow that.
Please comment.

### DIFF
--- a/cmd/gb/build.go
+++ b/cmd/gb/build.go
@@ -4,7 +4,6 @@ import (
 	"flag"
 	"fmt"
 	"go/build"
-	"path/filepath"
 	"time"
 
 	"github.com/constabulary/gb"
@@ -66,13 +65,6 @@ var BuildCmd = &Command{
 func resolvePackages(ctx *gb.Context, args ...string) ([]*gb.Package, error) {
 	var pkgs []*gb.Package
 	for _, arg := range args {
-		if arg == "." {
-			var err error
-			arg, err = filepath.Rel(ctx.Srcdirs()[0], projectroot)
-			if err != nil {
-				return pkgs, err
-			}
-		}
 		pkg, err := ctx.ResolvePackage(arg)
 		if err != nil {
 			if _, ok := err.(*build.NoGoError); ok {
--- a/cmd/gb/main.go
+++ b/cmd/gb/main.go
@@ -111,9 +111,9 @@ func main() {
 	}
 }
 
-// importPathsNoDotExpansion returns the import paths to use for the given
+// importPathsNoThreeDotExpansion returns the import paths to use for the given
 // command line, but it does no ... expansion.
-func importPathsNoDotExpansion(ctx *gb.Context, args []string) []string {
+func importPathsNoThreeDotExpansion(ctx *gb.Context, args []string) []string {
 	srcdir, _ := filepath.Rel(ctx.Srcdirs()[0], projectroot)
 	if srcdir == ".." {
 		srcdir = "."
@@ -142,7 +142,7 @@ func importPathsNoDotExpansion(ctx *gb.Context, args []string) []string {
 
 // importPaths returns the import paths to use for the given command line.
 func importPaths(ctx *gb.Context, args []string) []string {
-	args = importPathsNoDotExpansion(ctx, args)
+	args = importPathsNoThreeDotExpansion(ctx, args)
 	var out []string
 	for _, a := range args {
 		if strings.Contains(a, "...") {
--- a/cmd/gb/test.go
+++ b/cmd/gb/test.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"go/build"
-	"path/filepath"
 	"time"
 
 	"github.com/constabulary/gb"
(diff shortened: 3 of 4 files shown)


davecheney:
Thanks for your contribution. The problem with #48 is `gb.Build` and `gb.Test` take a `*gb.Package`, which is ok, but what is the name of a package from `$PROJECT/src/`. How would you import it ?

Can you explain a little more about your file structure in #48 so I can consider the outcome of applying this change.

mariokostelac:
I just realised that `gb build .` and `gb build /` both work, but `gb test .` or `gb test /` do not. I will update you as soon as I fix it.

davecheney:
Hmm, / should never be an acceptable path specifier. I don't think that is
a case we should support.

mariokostelac:
It worked out of the box, I do not know if it worked before my changes.
Could you please explain what was the purpose of https://github.com/mariokostelac/gb/commit/af2631924670e6b855824ec5ab2910b3409a29cd#diff-bb1dd15b956c336e94bbc3724b404711L69?

mariokostelac:
It took quite a time to get it right and to understand what's happening there, but it's finally here. The idea is to have a special name for that package ("." is already used by go build toolchain so I've used the same).
Since rest of the system already supported "." package (actually additional resolving for "." in build.gb and test.gb just broke that), only thing that we didn't have was to add "." package while getting all the packages.

Structure:
```
project/
  src/
    main.go
    repo/
      sth.go
```

`gb build .` in `project` builds main package.
`gb build .` in `project/src` builds main package.
`gb build ...` in `project` builds all the packages, including main.
`gb build .` in `project/src/repo` builds repo package.

Same applies to `gb test`.

davecheney:
Thank you very much for working on this. I deeply appreciate all the help I can get on this project.

Having said that, I am still conflicted about what building a package who's source is directly in `$PROJECT/src/`. By my reckoning that package has no name. This has some consequences, if the package has no name, then it cannot be cached. If the package has no name it cannot be imported into another package.

In your example
```
project/
  src/
    main.go
```
In this example this is is a command (ie, main.go starts with the line `package main`) then the name of the command is defined by the name of the directory it's in. But in this case, the command has not name, because of the rules above. 

Do you agree with this assessment ?

mariokostelac:
I agree, partly.
Statement related to caching is correct (of course, we can use _ or something similar for package name), but the one related to importing is not. Some packages are not created for being imported, like `main`, for example.

Currently the binary gets a name `bin`, which is completely fine with me.
Building the project with main source in `src/` is just more convenient for me, without extra layer of directories and very clean for smaller projects.

As biggest advantage of providing that is ramp up time for starters. When I run `gb build` for the first time, I was searching binaries around, reading the docs etc until I found out what happened there. It seemed buggy for me (which does not mean it is, maybe other people didn't fall into same problem).

davecheney:
Thanks for your patience. 

The more I look into this, the more uncomfortable I get with the idea of allowing this as a special case; I'm worried that what is a well meaning shortcut, just makes things more confusing.

Consider this example
```
lucky(/tmp/gopath/src) % ls
main.go
lucky(/tmp/gopath/src) % cat main.go 
package main

func main() {
        println("hello world")
}
lucky(/tmp/gopath/src) % env GOPATH=/tmp/gopath go list
_/tmp/gopath/src
lucky(/tmp/gopath/src) % env GOPATH=/tmp/gopath go install -v ./...
go install: no install location for directory /tmp/gopath/src outside GOPATH
lucky(/tmp/gopath/src) % env GOPATH=/tmp/gopath go build -v ./...                                                                                               
_/tmp/gopath/src
```

Looking at the behaviour of the go tool, `go list` can see the main.go file, but makes up a package name for it, `_/tmp/gopath/src`. 

`go build` can build the package, as long as you use the globing operator, but it ends up calling it `src`
```
lucky(/tmp/gopath/src) % ls -al
total 1044
drwxrwxr-x 2 dfc dfc      80 May 12 20:20 .
drwxrwxr-x 3 dfc dfc      60 May 12 20:10 ..
-rw-rw-r-- 1 dfc dfc      54 May 12 20:11 main.go
-rwxrwxr-x 1 dfc dfc 1064624 May 12 20:20 src
```
`go install` cannot install the package, it gives a confusing error message.

So, we can probably do better than the go tool here, but it would mean taking an opinionated stance on the name of the binary; `src` isn't acceptable. I'm sure this will be the source of a feature request to make it configurable.

But the bit I am concerned about it having to explain that this package is not importable but another package in the same project.

What I'd like to do is prohibit source files at the top level. Sure this makes things a little harder for beginners, but avoids bad habits ?

What do you think ?

mariokostelac:
Now I agree with you. Since current go tools do not support that special package very good, it seems that doing that on gb level would just introduce a new layer for bugs.

Related to issue with beginners, it seems reasonable to create a `gb init` command that would create
```
src/
  main/
vendor/
  src/
```

What do you think about that? First as a plugin.

davecheney:
I'm fine with this being a plugin, it sounds like an excellent idea.

I'll open a new issue for gb-init, and update this issue to explain that
we're not going to do this.

mariokostelac:
Thanks!

davecheney:
With your permission I'm closing this issue and will update #46 with the outcome

Issue #47, opened 2015-05-05
Only show compiler errors in normal output mode.


ardan-bkennedy:
When performing a normal build all I care about is seeing compiler messages. Everything else is noise.

% gb build
animals.go:17: syntax error: unexpected EOF, expecting } (edited)

Anything else should be provided only with a switch

%db build -v
2015/05/05 18:33:32 command "build" failed: [/Users/bill/go/pkg/tool/darwin_amd64/6g -p animals -pack -o /var/folders/8q/d2pfdk_x4qd4__l6gypvzsw40000gn/T/gb782637558/animals.a -I /var/folders/8q/d2pfdk_x4qd4__l6gypvzsw40000gn/T/gb782637558 -I /Users/bill/gb/test/pkg/darwin/amd64 -complete animals.go]: exit status 1
animals.go:17: syntax error: unexpected EOF, expecting }

PR #54, opened 2015-05-07
[RFC] reduce logging output in the default case

davecheney:
Updates #47 

Before:
```
% gb build .../sftp
2015/05/07 21:20:56 INFO project root "/home/dfc/devel/demo"
2015/05/07 21:20:56 INFO compile github.com/kr/fs
2015/05/07 21:20:56 INFO compile golang.org/x/crypto/ssh
2015/05/07 21:20:56 INFO install /home/dfc/devel/demo/pkg/linux/amd64/github.com/kr/fs.a
2015/05/07 21:20:57 INFO install /home/dfc/devel/demo/pkg/linux/amd64/golang.org/x/crypto/ssh.a
2015/05/07 21:20:57 INFO compile github.com/pkg/sftp
2015/05/07 21:20:57 INFO install /home/dfc/devel/demo/pkg/linux/amd64/github.com/pkg/sftp.a
2015/05/07 21:20:57 INFO build duration: 1.823214205s map[compile:1.854594245s]
```

After:
```
% gb build .../sftp                                                                                                                         
# compile github.com/kr/fs
# compile golang.org/x/crypto/ssh
# install /home/dfc/devel/demo/pkg/linux/amd64/github.com/kr/fs.a
# install /home/dfc/devel/demo/pkg/linux/amd64/golang.org/x/crypto/ssh.a
# compile github.com/pkg/sftp
# install /home/dfc/devel/demo/pkg/linux/amd64/github.com/pkg/sftp.a
```

This change reduces the amount of logging in the default case.

`-q` is still not the default, by default `gb` will still produce some output as it runs.

This is a deliberate decision based on watching new gophers use the `go` tool. Generally after replacing their Go install new gophers were confused that their builds suddenly take significantly longer (because the toolchain mismatch defeats incremental compilation). In those cases I wished that `go {build,install}` defaulted to `-v`, hence the motivation for `gb` being more talkative by default.

@wkennedy

### DIFF
--- a/cmd/gb/build.go
+++ b/cmd/gb/build.go
@@ -48,7 +48,7 @@ var BuildCmd = &Command{
 		ctx.Force = F
 		ctx.SkipInstall = FF
 		defer func() {
-			gb.Infof("build duration: %v %v", time.Since(t0), ctx.Statistics.String())
+			gb.Debugf("build duration: %v %v", time.Since(t0), ctx.Statistics.String())
 		}()
 
 		pkgs, err := resolvePackages(ctx, args...)
--- a/cmd/gb/main.go
+++ b/cmd/gb/main.go
@@ -95,7 +95,7 @@ func main() {
 	}
 	project := gb.NewProject(root)
 
-	gb.Infof("project root %q", project.Projectdir())
+	gb.Debugf("project root %q", project.Projectdir())
 
 	ctx, err := project.NewContext(
 		gb.GcToolchain(gb.Goroot(*goroot)),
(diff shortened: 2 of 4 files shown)


bketelsen:
yes please.  Too much output currently.

davecheney:
Thanks Brian. I want to keep working on this to get it down to one line per
package. Less than that will require the `-q` flag or some negotiation.

ChrisHines:
+1 to one line per package. In my use of the go tool these days I habitually run `go install -v` so I can see which packages are getting rebuilt. IMHO that is a good default for `gb`.

davecheney:
Thanks for the feedback.  I hear you all loud and clear.

davecheney:
I'm going to commit this now as an improvement to what exists currently. It is clear that people want even less output in the default mode, don't really care about the output in debug mode, and `-q` should of course shut up.

Issue #52, opened 2015-05-06
New subcommand, gb list


davecheney:
`go list` is a very useful command and `gb` could benefit from a similar tool to introspect a project.

`gb list` should probably be integrated into `gb`, but I would like to see it initially developed as a plugin to road test the plugin API (such that it is).

`gb list` should work the same as `go list`; you supply a `text/template` which is executed in an environment that has access to `*gb.Project`, `*gp.Context` and maybe a `[]*gb.Package` from the resolved package set.

This template can be supplied on the command line, similar to `go list -f`, but I'd also like to see it passed into the command via stdin, eg `echo "{{ .Context }}" | go list`

fatih:
This would be also helpful for editors. I'm using extensively `go list` to create the list of dependencies any many other things that I'm passing to `go build` or `go install`. So having it would make the integration for editors more easily.

freeformz:
+1 also for things like heroku-buildpack-go

mariokostelac:
@freeformz don't worry, already planning to do that :)

davecheney:
@mariokostelac #76 contains an initial cut of `gb-list`, feel free to take it and run with it.

paxan:
Anybody wants to try the buildpack... See #85

PR #77, opened 2015-05-12
gb-list

mariokostelac:
Added option to read from stdin. By default, format is `{{.ImportPath}}`, but we can redefine with
`gb list -f "format"` and
`echo "format" | gb list -s`

Connected with #52

### DIFF
--- a/cmd/gb-list/main.go
+++ b/cmd/gb-list/main.go
@@ -21,7 +21,7 @@ func main() {
 		format      string
 		formatStdin bool
 	)
-	flag.StringVar(&projectroot, "R", os.Getenv("GB_PROJECT_ROOT"), "set the project root")
+	flag.StringVar(&projectroot, "R", os.Getenv("GB_PROJECT_DIR"), "set the project root")
 	flag.StringVar(&format, "f", "{{.ImportPath}}\n", "format template")
 	flag.BoolVar(&formatStdin, "s", false, "read format from stdin")
 
--- a/cmd/gb-list/main.go
+++ b/cmd/gb-list/main.go
@@ -58,13 +58,15 @@ func main() {
 	}
 
 	if jsonOutput {
+		views := make([]*PackageView, 0, len(pkgs))
 		for _, pkg := range pkgs {
-			encoded, err := json.MarshalIndent(NewPackageView(pkg), " ", "  ")
-			if err != nil {
-				gb.Fatalf("Error occurred during json encoding: %v", err)
-			}
-			fmt.Println(string(encoded))
+			views = append(views, NewPackageView(pkg))
 		}
+		encoded, err := json.MarshalIndent(views, " ", "  ")
+		if err != nil {
+			gb.Fatalf("Error occurred during json encoding: %v", err)
+		}
+		fmt.Println(string(encoded))
 	} else {
 		tmpl, err := template.New("list").Parse(format)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


mariokostelac:
@davecheney Should we list all packages from vendor, too?

davecheney:
Good question. Let's not and see who complains.

mariokostelac:
Ok, what else do we need before merging?

cmd/gb-list/main.go
```diff
@@ -17,15 +20,21 @@ func main() {
 	var (
 		projectroot string
 		format      string
+		formatStdin bool
+		jsonOutput  bool
 	)
-	flag.StringVar(&projectroot, "R", os.Getenv("GB_PROJECT_ROOT"), "set the project root")
+	flag.StringVar(&projectroot, "R", os.Getenv("GB_PROJECT_DIR"), "set the project root")
 	flag.StringVar(&format, "f", "{{.ImportPath}}\n", "format template")
+	flag.BoolVar(&formatStdin, "s", false, "read format from stdin")
```
davecheney: I think we can make this more automatic, but this will do for a start.

cmd/gb-list/package_view.go
```diff
@@ -0,0 +1,27 @@
+package main
+
+import "github.com/constabulary/gb"
+
+type PackageView struct {
```
davecheney: Did this come from the Go sources ? If more than a little of it was copied, best to also copy the copywrite statement from the top of the file.

cmd/gb-list/main.go
```diff
@@ -47,9 +56,27 @@ func main() {
...
-	for _, pkg := range pkgs {
-		if err := tmpl.Execute(os.Stdout, pkg); err != nil {
-			gb.Fatalf("unable to execute template: %v", err)
+
+	if jsonOutput {
+		views := make([]*PackageView, 0, len(pkgs))
+		for _, pkg := range pkgs {
+			views = append(views, NewPackageView(pkg))
+		}
+		encoded, err := json.MarshalIndent(views, " ", "  ")
```
davecheney: Can you do this without buffering the string ? Ie construct a json.Encoded over os.Stdout

davecheney:
Pretty close and we can merge this as a first pass.

cmd/gb-list/main.go
mariokostelac: I agree

cmd/gb-list/package_view.go
mariokostelac: Nope, it is a partial view. We miss some fields like `Stale` and `Target`.

cmd/gb-list/main.go
mariokostelac: I will try again. I was not able to find a way to write prettified JSON with pure json encoder.

davecheney: Does it need to be indented ? I'm assuming the json form is only for
machines to consume, and they don't care about pretty printing.

How about this, use json.NewEncoder and don't indent. People that want it
indented can pipe it through jsonpp. We can always make this better later.

mariokostelac: If it is just for machines, indentation is not important, I agree.

davecheney: I forgot that go list even spoke json. If you hadn't already written that
code, I would have said not to work about it.

davecheney:
I've merged this so I can document it on getgb.io. The command also needs godoc package docs, and a HUGE warning to say that they gb.Package structure is not stable.

Issue #61, opened 2015-05-10
Goimports plugin


mattetti:
While I really like the direction gb is taking, the main obstacle to adoption for me is the fact that it breaks my tool chain, more specifically goimports which kicks in every time is save a go file inside vim.

davecheney:
yup, won't be hard, it'll end up looking like
```
gb goimports $IMPORTPATH # applies goimports to $IMPORTPATH
go goimprots # applies goimports to the whole of $PROJECT/src
```
Will that work with your workflow ?

mattetti:
@davecheney so I would have to invoke these 2 commands every time I save and before I call goimports?

davecheney:
@mattetti no, those are just to sample invocations, it'll just be one command.

mattetti:
yes, perfect

Xe:
I've ended up doing things like:

```sh
export GOPATH=$(pwd):$(pwd)/vendor
export PATH=$(pwd)/bin:$PATH
```

and sourcing that when I am in a gb project.

Xe:
Or even better, stick this in your `.lvimrc` for a project:

```vimscript
let $GOPATH = getcwd() . ":" . getcwd() . "/vendor"
```

davecheney:
That won't work if you are in any directory but the project room

ches:
@Xe: something like this might be more robust against Vim's working directory changing, as @davecheney pointed out:

```vim
" Guard multiple execution for currently edited file
if g:localvimrc_sourced_once_for_file | finish | endif

" Guard multiple execution for the running vim instance
if g:localvimrc_sourced_once | finish | endif

" -- Once per Vim instance --
let $GOPATH = g:localvimrc_script_dir . ':' . g:localvimrc_script_dir . '/vendor'
```

See `:help localvimrc-variables`.

vim-go master is now supposed to have some auto-`GOPATH` functionality when detecting a Godep or gb project, but it doesn't seem to work for me yet. See #42 for some links.

ches:
@davecheney: The gb-goimports plugin from #80 is still on a branch—is that pending work on pulling out some of the reusable code in that PR for other plugins to use, as mentioned on the thread, or some other undecided considerations for plugins?

davecheney:
I think that branch is dead, sadly, I don't think it would merge any more.

PR #80, opened 2015-05-13
gb-goimports plugin to run goimports within project scope

rynlbrwn:
See #61 

This plugin simply runs [`goimports`](http://golang.org/x/tools/cmd/goimports) with a `GOPATH` scoped to the project and its vendor directory. If it is invoked directly, rather than through `gb`, it runs `gb env` to find the project directory.

For emacs, I'm using `(setq gofmt-command "gb-goimports")`.

### DIFF
--- a/cmd/gb-goimports/main.go
+++ b/cmd/gb-goimports/main.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+func main() {
+	c := exec.Command("goimports", os.Args[1:]...)
+	c.Env = env()
+	c.Stdin = os.Stdin
+	c.Stdout = os.Stdout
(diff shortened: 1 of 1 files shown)


davecheney:
Thanks. I'll merge this into a branch and work on it some more. There is some shared code that can be pulled out of cmd/gb that would be useful for a plugin to be able to detect the project root.

rynlbrwn:
Okay. I was also thinking about removing the dependency of having `goimports` installed by just using the `imports` package that it uses. Depending on how much you want it to work with existing tools, you could have a different command line interface that way.

davecheney:
I'd like to see that happen.

I looked at the source for x/tools/cmd/goimports a few days ago and

a. a lot of it had been copied from gofmt (I think)
b. it just resolves packages and calls the imports package.

I think a wrapper is ok for now, but directly integrating with the import's
package would be better in the long run.

davecheney:
I've merged this into a branch, please work on a branch next time.

Issue #83, opened 2015-05-13
Import cycles not detetcted


markuslindenberg:
`gb` will hang when I have import cycles in my code, while go will detect this and refuse to build:

```
$ find .
.
./src
./src/bar
./src/bar/bar.go
./src/foo
./src/foo/foo.go

$ cat src/foo/foo.go 
package main

import _ "bar"

func main() {
}

$ cat src/bar/bar.go 
package bar

import "bar"

$ export GOPATH=$(pwd)
$ go install foo
import cycle not allowed
package foo
	imports bar
	imports bar
$ gb build foo
```

`gb` will hang with 100% CPU usage on one core.

mariokostelac:
Working on that.

davecheney:
Sorry about that, the code in loadPackage passes down a map to track import
loops, but it isn't hooked up yet.

markuslindenberg:
Yeah, thanks so much and thanks for `gb`, love it. :+1:

PR #84, opened 2015-05-13
Add import cycle detection

mariokostelac:
Since we do not use real stack, it does not seem obvious for me to detect a cycle of length longer than 2.

### DIFF
--- a/context.go
+++ b/context.go
@@ -79,7 +79,7 @@ func (c *Context) Pkgdir() string {
 
 // ResolvePackage resolves the package at path using the current context.
 func (c *Context) ResolvePackage(path string) (*Package, error) {
-	return c.loadPackage(make(map[string]bool), path)
+	return c.loadPackage(nil, path)
 }
 
 // ResolvePackageWithTests resolves the package at path using the current context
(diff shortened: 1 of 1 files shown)


mariokostelac:
Cycle reporting added

davecheney:
Thanks for working on this. I had originally intended for the map that is being passed down to be used for import cycle detection. If it's not used, then please remove it, a slice is probably the better solution.

Please also add a test.

mariokostelac:
No problem, I will add tests and remove original stack tomorrow. 
It will increase complexity of checking to `O(n)`, but people with very big dependency graph will have bigger problems than that.

kisielk:
I doubt any import chain will be long enough that checking a slice in O(n) is going to pose a problem...

davecheney:
That was why I added the map in the first place. Why not use the map, the
code is all there, just no check before recursing.

If a cycle is detected just return an error. We can say how the cycle was
formed, but we can do that later as an improvement.

mariokostelac:
I've added tests for different cycle lengths (to cover it correctly from algorithm perspective) so we have 1, 2, 4.
1 - edge case
4 - to be sure that we are not stuck on detecting just the cycle of length 2 (I can imagine people forgetting about cycles longer than 2)

I know that the way of testing if goroutine is not the nicest one, but using `time.AfterFunc` and asserting after a second did not work for me.

Also, if somebody know how to do error checking better, please teach me, I had to be sure that I am getting my error, not some other (like package not found).

davecheney:
Thanks for fix this.

Sorry, the test is not acceptable. Testing with time is unreliable. I think putting a panic if the stack size grows above say 100 would be a decent way of testing this.

Please drop the test and I'll merge this. Also edit the issue description and add the text

     Fixes #83

mariokostelac:
No problem, sir!

PR #88, opened 2015-05-14
fix gb-list behaviour inside src/%packagename

mariokostelac:
Wrong path had been passed. In ImportPaths function it is clear that second argument is cwd.


davecheney:
Thanks! This diff is dirty, there are bits of gb-init in it.

mariokostelac:
Sorry, going to fix that.

PR #89, opened 2015-05-14
Import cycle tests

mariokostelac:


context_test.go
```diff
@@ -0,0 +1,34 @@
+package gb
+
+import (
+	"runtime/debug"
+	"strings"
+	"testing"
+)
+
+func testImportCycle(pkg string, t *testing.T) {
```
davecheney: you can reuse the table in `build_test.go`, 

add entries for "cycle0", etc, and include the error you expect.

That should be good enough

mariokostelac: Ok, but will leave stack limit there. Just in case if we break it in the future so we don't lose our heads on finding the issue.
Sounds good?

davecheney: Oh, I didn't even see that. I'll merge this as is, it'll get messy to hammer it into the general test.

mariokostelac:
thanks

davecheney:
No. Thank you, valued associate.

PR #90, opened 2015-05-14
Include error

freeformz:
I ran across a problem today where I needed this info to understand why
gb wasn't building anything. Turns out it was a misvendored dependency.


davecheney:
LGTM. This current logic is a side effect of reusing the `go/build` loader. I hope to write a new one in the next few weeks which will be better suited to the gb project format.

PR #103, opened 2015-05-17
Initial cgo support

davecheney:
Fixes #92 

Initial cgo support, many things are missing, but basic cgo compilation works.

PR #106, opened 2015-05-18
runOut: don't lose command output to stdout

cryptix:
Hi,

during my hackish attempt of adding gb to [gin](https://github.com/codegangsta/gin), I noticed that the command output of the `runOut` method is only printed to stdout in the error case.

This means that information and output of the compiler and other commands can't be read by tools using gb by themselfs. This PR adds a simple error type that captures the original error, the command args and their output. 

Ps: see PR codegangsta/gin#73 if you want to play with it.

### DIFF
--- a/gb.go
+++ b/gb.go
@@ -59,9 +59,9 @@ func run(dir, command string, args ...string) error {
 	var buf bytes.Buffer
 	err := runOut(&buf, dir, command, args...)
 	if err != nil {
-		fmt.Printf("# %s %s\n%s", command, strings.Join(args, " "), buf.String())
+		return fmt.Errorf("# %s %s: %v\n%s", command, strings.Join(args, " "), err, buf.String())
 	}
-	return err
+	return nil
 }
 
 func runOut(output io.Writer, dir, command string, args ...string) error {


davecheney:
Thanks. This is fine but drop the output field, there is no standard way of printing an io.Writer.

cryptix:
> drop the output field,

Hrm.. the juicy information (`xyz is undefined`, `can't use foo as bar in assignment..`) is in the command output. You only get `exit status 1` without it.

>  there is no standard way of printing an io.Writer.

This was the most confusing thing to me when introducing the output field as string, I didn't notice is was a writer because it was printed like that.

Since only `run()` uses `runOut()` and it also didn't look at the passed buffers content afterwards, I dropped the `run()` wrapper and converted the `exec.Command` to use `CominedOutput()` instead of `Run()`. The output is only captured in the error case.

How does this look to you?

davecheney:
Have a look at the changes to this file landing in #103. I can break out
the changes to run/runOut if you need.

cryptix:
> I can break out the changes to run/runOut if you need.

Oh, don't bother. I'm happy to wait and rebase my changes ontop once cgo is merged. (yeeey! :smile:)

gb.go
```diff
@@ -55,23 +54,29 @@ func copyfile(dst, src string) error {
...
+	return s
+}
+
+func run(dir, command string, args ...string) error {
 	cmd := exec.Command(command, args...)
 	cmd.Dir = dir
-	cmd.Stdout = output
-	cmd.Stderr = os.Stderr
-	// Debugf("cd %s; %s", cmd.Dir, cmd.Args)
-	fmt.Println(strings.Join(cmd.Args, " "))
```
cryptix: This `fmt.Println` sneaked in from the cgo changes, I think. I changed it back to `Debugf` which seems more in line with the _be less noisy_ discussion.

davecheney: Already fixed.

cryptix:
Aaah, `libgcc()` actually uses `runOut()` to output the generated content.. I'll adjust it accordingly.

cryptix:
Okay, the new custom error is now returned from `run()` and `runOut()` stays unchanged (apart from the `Debugf()` change I commented on above).

cryptix:
If you are okay with this change, I'd squash them to spare the history from this back and forth.

One thing still irks me, some output is still tunneled to `os.Stderr` directly. Some tool might leave the interesting information there but this is only hypothetical.

davecheney:
Do you want to continue with this PR ?

cryptix:
> Do you want to continue with this PR ?

Yes, please! Did I miss something?

davecheney:
Wercker says no at the moment, do you want to merge master?

cryptix:
Okay - rebased on master and made one commit out of it.

davecheney:
I'm not sure this really adds much. This PR looses the command being executed, it just records the args. Does what exists today not do what you need ?

cryptix:
My problem is that when other tools use gb's code directly (not shelling out to the gb tool, like `os/exec`) their error output are lost to stdout and can't be reused.

In gin, I just wanted to output the errors to the browser as well. Other tools might want to process the source locations of the errors. The latter would still be possible through shelling out but If you already use gb as a package to find the project root and parse the packages in it, it feels weird to go back to `os/exec` for building a package.

cryptix:
>  This PR looses the command being execute

Whops.. Yes, sorry about that. It now does capture the command, too.

gb.go
```diff
@@ -55,13 +55,28 @@ func copyfile(dst, src string) error {
...
+	s += e.Output
+	return s
+}
+
 func run(dir, command string, args ...string) error {
 	var buf bytes.Buffer
 	err := runOut(&buf, dir, command, args...)
 	if err != nil {
-		fmt.Printf("# %s %s\n%s", command, strings.Join(args, " "), buf.String())
+		return RunErr{Err: err, Cmd: command, Args: args, Output: buf.String()}
```
davecheney: why can't this just be

err = fmt.Errorf("# %s %s: %v\n%s", command, strings.Join(args, " "), err, buf.String())

cryptix: Agreed, much simpler. The extra type really didn't add any extra value.

davecheney:
Thanks for sticking with me on this.

cryptix:
Thanks for pushing me to a better optimum.

PR #124, opened 2015-05-24
Improved testflag parsing

demizer:
Last of the testflag stuff.

Copied the flag spec object from the [upstream testflag.go](https://github.com/golang/go/blob/master/src/cmd/go/testflag.go), this removes the need for looping through flag.FlagSet and cleans up the code and tests a bit.

Many more test cases were added.

### DIFF
--- a/cmd/gb/test.go
+++ b/cmd/gb/test.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"flag"
+	"fmt"
+	"os"
 	"time"
 
 	"github.com/constabulary/gb"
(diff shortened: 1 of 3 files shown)


cmd/gb/test.go
```diff
@@ -13,13 +15,19 @@ func init() {
...
+	testProfile   bool
+	testCover     bool
+	testCoverMode string
+	testCoverPkg  string
 )
 
 func addTestFlags(fs *flag.FlagSet) {
 	addBuildFlags(fs)
 	fs.BoolVar(&testCover, "cover", false, "enable coverage analysis")
+	fs.StringVar(&testCoverMode, "covermode", "set",
```
davecheney: no need to wrap this, Go isn't python, we don't have 78 col hard limits

cmd/testflag.go
```diff
@@ -1,59 +1,175 @@
...
+// testFlagSpec defines a flag we know about.
+type testFlagSpec struct {
+	boolVar    bool // True if the flag is type bool
+	passToTest bool // pass to Test
+	passToAll  bool // pass to test plugin and test binary
+	present    bool // The flag has been seen
+}
+
+// testFlagDefn is the set of flags we process.
+var testFlagDefn = map[string]*testFlagSpec{
```
davecheney: does this need to be `*testFlagSpec`, can it just be a value type ?

davecheney:
Lovely. Just a few nit picks, but this is looking great.

cmd/testflag.go
demizer: I tried using a value type, but with Go, you can't directly set a struct variable like ```testFlagDefn["flag"].present = true``` unless it is a reference type (see https://github.com/golang/go/issues/3117). I can change this behavior to use the workaround described in the issue if you wish, or change it to something else. FWIW, upstream uses the same technique: https://github.com/golang/go/blob/master/src/cmd/go/testflag.go#L249

davecheney: Nah, that's fine, that make sense, please leave it as it is.

PR #126, opened 2015-05-25
[RFC] a new version of gb-vendor

davecheney:
Updates #17 
Updates #29 
Updates #71 
Updates constabulary/gb-vendor#1
Updates constabulary/gb-vendor#2
Updates constabulary/gb-vendor#4
Updates constabulary/gb-vendor#5
Updates constabulary/gb-vendor#6
Updates constabulary/gb-vendor#7

Closes #130

This PR contains a new version of the `gb-vendor` plugin. The aim is to replace, in short order, the existing constabulary/gb-vendor plugin, which will then be removed.

This new `gb-vendor` does not rely on `go get`. 

## Usage

`gb-vendor` accepts subcommands. At the moment there are thee subcommand is `fetch`, `update` and `delete` (thanks to @davidkaya).
```
% gb vendor
Usage:
  gb delete [flags] [package] - deletes a local dependency
  gb fetch [flags] [package] - fetch a remote dependency
  gb update [flags] [package] - updates a local dependency
```
 `fetch` takes a single argument, an import path. Here is an example of using `gb vendor fetch`
```
% gb vendor fetch github.com/kr/fs
Cloning into '/tmp/gb-vendor-161722046'...
remote: Counting objects: 18, done.
remote: Total 18 (delta 0), reused 0 (delta 0), pack-reused 18
Unpacking objects: 100% (18/18), done.
Checking connectivity... done.
copyfile(dst: /home/dfc/devel/demo4/vendor/src/github.com/kr/fs/LICENSE, src: /tmp/gb-vendor-161722046/LICENSE)
copyfile(dst: /home/dfc/devel/demo4/vendor/src/github.com/kr/fs/Readme, src: /tmp/gb-vendor-161722046/Readme)
copyfile(dst: /home/dfc/devel/demo4/vendor/src/github.com/kr/fs/example_test.go, src: /tmp/gb-vendor-161722046/example_test.go)
copyfile(dst: /home/dfc/devel/demo4/vendor/src/github.com/kr/fs/filesystem.go, src: /tmp/gb-vendor-161722046/filesystem.go)
copyfile(dst: /home/dfc/devel/demo4/vendor/src/github.com/kr/fs/walk.go, src: /tmp/gb-vendor-161722046/walk.go)
copyfile(dst: /home/dfc/devel/demo4/vendor/src/github.com/kr/fs/walk_test.go, src: /tmp/gb-vendor-161722046/walk_test.go)
```
As usual the output is verbose, it will be toned down in later revisions.

[...]

PR #127, opened 2015-05-25
Added support for update and delete in new gb-vendor

davidkaya:
Updates #126

I have added support for update and delete.

There are a lot of possible improvements (automatically detecting unused packages and deleting them with something like `gb vendor delete --unused`, or updating all dependencies `gb vendor update --all`, etc.). If you don't decline the PR (permanently) I will continue with adding other features.

Since the `gb vendor fetch` removes `.git` folder, it is required to remove old copy and clone new one.
I am open for improvements! :)

### DIFF
--- a/cmd/gb-vendor/vendor/manifest.go
+++ b/cmd/gb-vendor/vendor/manifest.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"reflect"
 )
 
 // gb-vendor manifest support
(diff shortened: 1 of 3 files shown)


cmd/gb-vendor/vendor/manifest.go
```diff
@@ -28,6 +29,18 @@ func (m *Manifest) AddDependency(dep Dependency) error {
 	return nil
 }
 
+// RemoveDependency removes a Dependency from the current Manifest
```
davecheney: full stop at the end of these lines (I know I'm not that consistent with this, but I should get better).

davecheney:
Thanks for this. I'm pleased how easy it was to add these two operations. I won't land this till #127 lands, so get out your brush and start painting that bike shed.

davecheney:
Oh, i see this is a PR on my branch. I've not used that form much. I guess I can merge it in then update #126. I'll try that now.

PR #134, opened 2015-05-26
Support for --all, --branch, --revision in gb vendor fetch and --all in gb vendor update

### LINKED ISSUE #137: gb vendor help text does not include the name of the plugin in the usage text.
% gb vendor
Usage:
  gb delete [flags] [package] - deletes a local dependency
  gb fetch [flags] [package] - fetch a remote dependency
  gb update [flags] [package] - updates a local dependency

Flags:

davidkaya:
Updates #126 
Implements (only `gb vendor fetch` for now) #136
Fixes #137 

Added support for `--all`, `--branch`, `--revision` flag in `gb vendor fetch`.
Added support for `--all` flag in `gb vendor update`:
* I think `gb vendor update` should not have `--branch` and `--revision` flag, since it should get those from Manifest file and pull last revision of the branch specified in the Manifest file (explained in detail in  [#136](https://github.com/constabulary/gb/issues/136#issuecomment-105551838)).

Possible issue:
~~* after calling `gb vendor delete <package>`, package's parent directory stays empty~~ Fixed

**NOTES:**
* Checking out revision (commit sha) places git repository into Detached HEAD -> branch name becomes *HEAD* (also in manifest file)

### DIFF
--- a/cmd/gb-vendor/delete.go
+++ b/cmd/gb-vendor/delete.go
@@ -36,7 +36,7 @@ var DeleteCmd = &cmd.Command{
 
 		m, err := vendor.ReadManifest(manifestFile(ctx))
 		if err != nil {
-			return fmt.Errorf("could not load manifest: %T %v", err, err)
+			return fmt.Errorf("could not load manifest: %v", err)
 		}
 
 		var dependencies []vendor.Dependency
@@ -47,7 +47,7 @@ var DeleteCmd = &cmd.Command{
 			p := args[0]
 			dependency, err := m.GetDependencyForImportpath(p)
 			if err != nil {
-				return fmt.Errorf("could not get dependency: %T %v", err, err)
+				return fmt.Errorf("could not get dependency: %v", err)
 			}
 			dependencies = append(dependencies, dependency)
 		}
@@ -56,15 +56,14 @@ var DeleteCmd = &cmd.Command{
 			path := d.Importpath
 
 			if err := m.RemoveDependency(d); err != nil {
-				return fmt.Errorf("dependency could not be deleted: %T %v", err, err)
+				return fmt.Errorf("dependency could not be deleted: %v", err)
 			}
 
 			localClone := vendor.GitClone{
 				Path: filepath.Join(ctx.Projectdir(), "vendor", "src", path),
 			}
-
 			if err := localClone.Destroy(); err != nil {
-				return fmt.Errorf("dependency could not be deleted: %T %v", err, err)
+				return fmt.Errorf("dependency could not be deleted: %v", err)
 			}
 		}
 		return vendor.WriteManifest(manifestFile(ctx), m)
--- a/cmd/gb-vendor/delete.go
+++ b/cmd/gb-vendor/delete.go
@@ -22,7 +22,7 @@ func init() {
 }
 
 func addDeleteFlags(fs *flag.FlagSet) {
-	fs.BoolVar(&deleteAll, "all", false, "update all dependencies")
+	fs.BoolVar(&deleteAll, "all", false, "delete all dependencies")
 }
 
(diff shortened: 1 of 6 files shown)


cmd/gb-vendor/delete.go
```diff
@@ -9,41 +10,69 @@ import (
...
+			path := d.Importpath
 
-		localClone := vendor.GitClone{
-			Path: filepath.Join(ctx.Projectdir(), "vendor", "src", path),
-		}
-		err = localClone.Destroy()
-		if err != nil {
-			return fmt.Errorf("dependency could not be deleted: %T %v", err, err)
-		}
+			if err != nil {
```
davecheney: i think this is dead code

cmd/gb-vendor/delete.go
```diff
@@ -9,41 +10,69 @@ import (
...
+				return fmt.Errorf("could not get dependency: %T %v", err, err)
+			}
+
+			err = m.RemoveDependency(d)
```
davecheney: consider using

if err := ... ; err != nil { 
   ...
}

cmd/gb-vendor/delete.go
```diff
@@ -9,41 +10,69 @@ import (
...
+			if err != nil {
+				return fmt.Errorf("dependency could not be deleted: %T %v", err, err)
```
davecheney: why print the error twice ?

davidkaya: @davecheney 
It is `%T %v`, I used the same format as you used in the `fetch` command so the error messages won't differ.

davidkaya:
Fixed issues that you have found in the `delete.go`.

cmd/gb-vendor/delete.go
davecheney: opps, that looks like that's left over debugging. Could you fix it for me ?

davidkaya: Yep sure, I will fix all double error occurences.

cmd/gb-vendor/fetch.go
```diff
@@ -70,4 +99,5 @@ var FetchCmd = &cmd.Command{
 		}
 		return wc.Destroy()
```
davidkaya: @davecheney I've noticed that you are returning wc.Destroy() here. Wouldn't it be more intuitive if you return vendor.WriteManifest(...)? Since the wc.Destroy() is not the main objective of the command. Atleast I've used that approach in the update and delete command.

davecheney: Yup, that will work, setup a defer to wc.Destroy then return
vendor.WriteManifest.

Also, there is no need to pack all these changes into one PR.

If you send me small PRs, I can review those faster and get them merged for
you.

davidkaya: Ok, I will create smaller PRs for feature changes

davecheney: Cool. I can see the direction your going so there is no need for extended
debate in PRs. If we stick to designing features up front in issues then
small PRs will work out fine.

Thanks

Dave

sundarv85:
Hello,

```
# gb vendor fetch --all
flag provided but not defined: -all
```

Is there an option in `vendor fetch` to fetch all the failed packages automatically?

Thanks
Sundar.

davecheney:
@sundarv85 please do not comment on old closed issues. It was decided not to add gb vendor fetch --all because, while convenient, it would encourage users to _not_ vendor their dependencies, instead rely on gb vendor as some replacement for go get.

sundarv85:
@davecheney Thanks for the clarification. That was missing in the issue, hence thought of asking it here itself. Next time, I will write in the groups or open a new issue.

PR #147, opened 2015-05-27
WriteManifest removes Manifest if there are 0 dependencies

### LINKED ISSUE #146: gb-vendor: writing a manifest file with 0 dependencies should remove the manifest
#134 added `gb vendor delete --all` which removes all deps and their records from the manifest file. 

I think it would be useful if WriteManifest detected that there were zero dependencies and removed the manifest instead, so you don't get
```
% cat ~/vendor/manifest 
{
	"version": 0,
	"dependencies": []
}
```

davidkaya:
Fixes #146

### DIFF
--- a/cmd/gb-vendor/vendor/manifest.go
+++ b/cmd/gb-vendor/vendor/manifest.go
@@ -94,13 +94,11 @@ type Dependency struct {
 // destroying a working vendorfile.
 func WriteManifest(path string, m *Manifest) error {
 	if len(m.Dependencies) == 0 {
-		// Check if files exists.
-		_, err := os.Stat(path)
-		if err != nil {
-			// Return nil because it is OK if the file does not exist yet.
-			return nil
+		err := os.Remove(path)
+		if !os.IsNotExist(err) {
+			return err
 		}
-		return os.RemoveAll(path)
+		return nil
 	}
 
 	f, err := os.Create(path)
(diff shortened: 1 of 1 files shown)


cmd/gb-vendor/vendor/manifest.go
```diff
@@ -89,9 +89,20 @@ type Dependency struct {
...
 // WriteManifest writes a Manifest to the path. If the manifest does
 // not exist, it is created. If it does exist, it will be overwritten.
+// If the manifest file is empty (0 dependencies) it will be deleted.
 // TODO(dfc) write to temporary file and move atomically to avoid
 // destroying a working vendorfile.
 func WriteManifest(path string, m *Manifest) error {
+	if len(m.Dependencies) == 0 {
+		// Check if files exists.
+		_, err := os.Stat(path)
+		if err != nil {
```
davecheney: this might be better written as

err := os.Remove(path) // not removeall, we only want to remove a file, not a directory by mistake
if !os.IsNotExist(err) {
     return err // some other kind of error
}
return nil // not found is ok

davidkaya: Yep, you are right, that is far more elegant. Fixed

davecheney:
LGTM. You should have permission to merge this now.

davidkaya:
It does not seem that I can merge it, it says that it can by merged by project collaborators but underneath there is this line: "Only those with write access to this repository can merge pull requests."

davecheney:
Please accept the invitation that github will have sent you.

davidkaya:
Oh, accepted. Maybe Github could show some notification on the website too.

PR #154, opened 2015-05-29
Add Command.ParseArgs

davecheney:
Currently the command handling logic will interpret arguments to
all comamnds as import paths unless the name of the command is `plugin`,
then it won't.

Make this more flexible by removing the hard coded name and replacing it
with a function on the Command structure with a reasonable default.


davecheney:
This is useful for commands like `gb imports` (clone of goimports) and `gb doc` which are not plugins, but need to handle their own arguments.

PR #156, opened 2015-05-29
cmd/gb: add gb doc

### LINKED ISSUE #74: Support 'gb doc'
Having 'gb doc' would actually be a fairly nice feature, because the docs would only be the project and the libraries the project uses, free from the pollution of stuff installed with go get. 

@davecheney "That can be implemented as a plugin pretty easily (godoc and go doc are effectively plugins to the go tool), please raise an issue so I don't forget."

davecheney:
Fixes #74 

gb doc is a simple wrapper around go doc / godoc. Invocation is the same

    gb doc $PACKAGE
    gb doc $PACKAGE $SYMBOL

Without arguments gb doc behaves as if you had passed `gb doc .` which
expands to the current working directory.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -27,6 +27,11 @@ type Command struct {
 
 	// Allow plugins to modify arguments
 	FlagParse func(fs *flag.FlagSet, args []string) error
+
+	// ParseArgs provides an alterntive method to parse arguments.
+	// By default, arguments will be parsed as import paths with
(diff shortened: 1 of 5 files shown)


davidkaya:
The PR contains the commit from #154, is it ok?

davecheney:
> The PR contains the commit from #154, is it ok?

Yeah, I haven't figured out a good way to do stacked branches with git yet.

Issue #135, opened 2015-05-26
gb-vendor: add purge subcommand that removes imports that are not referenced from $PROJECT/src


davecheney:
Closes constabulary/gb-vendor#7

davidkaya:
Suggestion:
We could parse imports from $PROJECT/src with https://golang.org/pkg/go/ast/ into Set (`map[string]struct{}`?) and compare it with imports from manifest file and remove those which are not referenced?

pires:
+1

davidkaya:
Are imports from the project already parsed and saved somewhere or do I have to parse the imports from $root/src? I wasn't able to find anything related to that.

davecheney:
This will be quite an involved feature as it requires loading all the source in `$PROJECT/src`, extacting the set of dependencies that are not met from the source iteslf; thus met from vendor/src/, then loading all the importpaths determinting packages references and unrefernced, then removing the unreferenced ones.

davidkaya:
Can I rely on the fact that all the dependencies (referenced and unreferenced) will be in the Manifest file or is it also required to parse the {{vendor/*}} structure? 

I think I should be able to rely on that, because the {{purge}} command is sub-command of {{gb vendor}}, thus the dependencies should be in the Manifest file.

davecheney:
yes, purge should only remove items mentioned in the manifest.

Issue #32, opened 2015-05-02
gb should support tranisitively building standard library packages


davecheney:
Related to #31 

`gb` currently ignores any package it thinks is part of the standard library; it just assumes they are up to date.

This may not be correct in a cross compilation world, where package resolution and `isStale` calculation could reach all the way back to `runtime`

Can will need to be taken that package resolution always goes, `$GOROOT`, `$PROJECT/src`, $PROJECT/vendor/src` -- it should never be possible for `$PROJECT/src/errors` for example, to override `$GOROOT/src/errors`.

However, when it comes to package compilation time, we want to cache the results of compiling source from `$GOROOT/src` in `$PROJECT/pkg` and this will be tricky.

tianon:
Just ran into this with the innocuous `cmd` name in https://github.com/tianon/go-tags.

`go build` happily spits out a binary, but `gb build` returns the wonderfully cryptic `FATAL command "build" failed: failed to resolve import path "cmd": no buildable Go source files in /usr/lib/go/src/cmd` (because `cmd` exists in `GOROOT`, so it must be `GOROOT/src/cmd` I want to compile, right?  except there are no `.go` files there so it errors out instead :smile:).

davecheney:
@tianon kicking this to a new issue.

tianon:
:heart:!

PR #159, opened 2015-05-29
Added purge sub-command to gb vendor command

davidkaya:
Implements #135


davidkaya:
I am not sure with one thing, and that is the {{stdlib}} map. It is already declared in {{gb}} package but unexported, so is it ok declaring it again here?

davecheney:
It's fine to export the Stdlib map from the `gb` package. It is only unexported because there were no other consumers, but now there are, so that is fine. In the future it is likely the Stdlib map will move out of the `gb` package when #32 is resolved.

davidkaya:
Done, I have exported it and changed `gb` to use exported one.

cmd/gb-vendor/purge.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/constabulary/gb/cmd/gb-vendor/vendor"
+)
+
+func init() {
+	registerCommand(PurgeCmd)
+}
+
+func parseImports(root string) (map[string]struct{}, error) {
+	var found = struct{}{}            // Does not take any space
+	pkgs := make(map[string]struct{}) // Set
```
davecheney: Map values always take at least one byte for the key, so map[string]bool makes a better set.

You also get to do 

if m["thing"] { // found }

rather than

if _, found := m["thing"]; found { ... }

cmd/gb-vendor/purge.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		fs := token.NewFileSet()
+		f, err := parser.ParseFile(fs, path, nil, parser.ImportsOnly)
+		if err != nil {
+			return err
+		}
+
+		for _, s := range f.Imports {
+			if _, ok := gb.Stdlib[s.Path.Value]; !ok {
```
davecheney: if !gb.Stdlib[...] { ... } 

see above

cmd/gb-vendor/purge.go
```diff
@@ -0,0 +1,92 @@
...
+				pkgs[strings.Replace(s.Path.Value, "\"", "", -1)] = found
+			}
+		}
+		return nil
+	}
+
+	if err := filepath.Walk(root, walkFn); err != nil {
```
davecheney: return filepath.Walk(...)

davecheney:
Some little nit picky comments. I haven't looked at this PR in detail yet; i'm amazed that it could be this simple.

cmd/gb-vendor/purge.go
davidkaya: If I `return pkgs, filepath.Walk(..)`, will the filepath.Walk(..) fill the `pkgs` correctly?

cmd/gb-vendor/purge.go
davidkaya: Would't the `map[string]bool` require one more byte for the value also?

davecheney: Maps always use at least one byte for the value.

cmd/gb-vendor/purge.go
davecheney: > On 31 May 2015, at 7:17 pm, Dávid Kaya <notifications@github.com> wrote:
> 
> In cmd/gb-vendor/purge.go <https://github.com/constabulary/gb/pull/159#discussion_r31388470>:
> 
> > +
> > +		fs := token.NewFileSet()
> > +		f, err := parser.ParseFile(fs, path, nil, parser.ImportsOnly)
> > +		if err != nil {
> > +			return err
> > +		}
> > +
> > +		for _, s := range f.Imports {
> > +			if _, ok := gb.Stdlib[s.Path.Value]; !ok {
> > +				pkgs[strings.Replace(s.Path.Value, "\"", "", -1)] = found
> > +			}
> > +		}
> > +		return nil
> > +	}
> > +
> > +	if err := filepath.Walk(root, walkFn); err != nil {
> If I return pkgs, filepath.Walk(..), will the filepath.Walk(..) fill the pkgs correctly?
> 
> 

good point, the order of evaluation is not defined, do

err := filepath.Walk(…)
return pkgs, err

The point is there is no point in checking if the error is nil or not, the caller will do that for us

cmd/gb-vendor/purge.go
davidkaya: Oh, ok. I thought they don't.

davidkaya:
Changed the map value type and removed that error check.

cmd/gb-vendor/purge.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			return fmt.Errorf("import could not be parsed: %v", err)
+		}
+
+		dependencies := make([]vendor.Dependency, len(m.Dependencies))
+		copy(dependencies, m.Dependencies)
+
+		for _, d := range dependencies {
+			if !imports[d.Importpath] {
+				dep, err := m.GetDependencyForImportpath(d.Importpath)
```
davecheney: Here is the case that I am concerned about

project/src/a/a.go imports
project/vendor/src/b/c/c.go
and 
projec/vendor/src/b/c/c.go imports
project/vendor/src/b

I can see two problems which this code as written. GetDependencyForImportPath won't match b/c. as b/ was the import path vendored. 

And if it did, then does the code remove b/c. or just b, or both ?

davidkaya: If I understood you correctly then something like this could work?
```go
func hasImportWithPrefix(d string) bool {
    for i := range imports {
        if strings.HasPrefix(i, d)
	    return true
        }
    }
    return false
}

for _, d := range dependencies {
    if !hasImportWithPrefix(imports, d.Importpath){
        dep, err := m.GetDependencyForImportpath(d.Importpath)
        ....
    }    
}
```

davecheney:
That'll probably work. 

I'm happy to see this land as is, it will work for most use case.

FWIW I think the 100% solution will require loading each source package, assigning it to some data structure, I'll call a source directory (although that is a little confusing), think "independent location of source code"; $PROJECT/src is one, and each entry inside $PROJECT/vendor/src{each manifest entry} are additional ones.

Once that is done, starting from each source package in $PROJECT/src, walk the dependency graph, and mark each source directory encountered. When that is done, any source directory that was not marked is unreachable and can be purged.

davidkaya:
Updated PR.

davecheney:
Sorry, you'll have to merge #169

davidkaya:
It should be ok now.

davecheney:
Sorry, you'll also have to integrate with #170

davecheney:
LGTM once you've merged.

davidkaya:
Please check last commits, if ok I will merge it.

davecheney:
LGTM.

Issue #152, opened 2015-05-29
stop duplicating documentation


davecheney:
Currently gb commands and plugins have usage messages, godoc, and are documented in getgb.io.

I want to reduce the number of times this documentation is repeated, specifically changing getgb.io to link to the godoc. There is no reason gb and it's plugins couldn't have godoc like this, https://godoc.org/rsc.io/grepdiff

davidkaya:
I would suggest changing only the getgb.io documentation into the godoc as you said but I would leave usage messages as they are now (and maybe use KISS principle to avoid over-documenting usage) because I guess user won't want to visit godoc/generate godoc just to see how to use a command.

davecheney:
At the moment I think il add pages to the site, the overwrite them with a redirect at the nginx level

PR #169, opened 2015-06-01
Refactor help, godoc, and package doc for cmd/gb

davecheney:
Updates #152 

Use go generate to produce command godoc for cmd/gb. getgb.io will be updated to point the usage page to the godoc.org page for cmd/gb


cmd/gb/env.go
```diff
@@ -13,8 +13,12 @@ func init() {
...
 var EnvCmd = &cmd.Command{
 	Name:      "env",
-	ShortDesc: "env prints the project environment variables",
-	Run:       env,
+	UsageLine: `env`,
+	Short:     "print project environment variables",
+	Long: `Env prints project environment variables.
+
+`,
+	Run: env,
```
davidkaya: @davecheney 
Is this formatting correct?

davecheney: Do you mean the extra \n's ? 

Yeah, the conversion to godoc and putting enough \n's to make sure lines are treated as headings requires them. Or at least is very flakey without them.

davidkaya: Oh ok.

davecheney: Yeah. It's not awesome, but it sort of works for now and closes #152.

PR #172, opened 2015-06-01
cmd/gb-vendor: add support for bitbucket and hg repos

davecheney:


davidkaya:
Looks good, I don't exactly know how `hg` commands work, so can't verify that.

davecheney:
The only bitbucket repos I can find are in .git format anyway.

This gives us the two main repo types that are driven by vanity imports. I'm going to move on to that next.

davidkaya:
What about adding some "pinging" feature? Checking if repo is git (`git ls-remote`), mercurial (`hg indetify`), etc. We could ping the repo path before cloning.

davecheney:
Maybe. How is that different from trying to do a clone and giving up if
there is an error ?

davidkaya:
Well I guess it won't differ but, at least in my opinion it would be cleaner. And we could reuse the ping for checking if a repository contains a branch that user wants to checkout. 
When fetching specific branch, it works now this way:
Clone repository -> Try to checkout branch -> If it fails (branch does not exist) then the repository stays cloned (maybe should be reported as bug)

With pinging:
Ping if repo has the branch -> If yes, then clone && checkout

davecheney:
ok. I'll merge this now.

I agree with your comments. I suspect this code will change when I implement the logic for vanity url probing.

I want to move onto that because it's blocking people trying to vendor golang.org/x/ repos

davidkaya:
Ok sure, no problem.

PR #173, opened 2015-06-01
Fix gb generate to use the arguments it's been given

prep:


prep:
This allows a "gb generate" to be run from the project directory.

davecheney:
LGTM. I didn't realise go generate took arguments, I thought it always ran in the CWD.

PR #177, opened 2015-06-01
Add support for gopkg.in imports

jamal:
I don't necessarily like that this needs to be supported, but since some existing libraries use the gopkg.in import path, I have to be able to import them as such in my vendor dir. Since gopkg.in acts as a git mirror, we can treat this just like a git repo.

I figured I'd make a PR in case this should be included, but I won't be offended if the PR just gets closed :) (I'll keep the change locally and hope that one day these libraries stop using gopkg.in).

In any case, this should work just like the other formats:

    gb vendor fetch gopkg.in/pg.v3

to fetch gopkg.in/pg.v3.


davecheney:
Thanks for this patch but I'd like to solve this in a more general manner, that is looking up the metadata from the vanity import. This is tracked in a few issues, but none we assigned, so I have created #178 and will merge the others into that one.

PR #179, opened 2015-06-02
cmd/gb-vendor/vendor: add string set operations

davecheney:


cmd/gb-vendor/vendor/stringset.go
```diff
@@ -0,0 +1,49 @@
...
+	return r
+}
+
+// contains returns true if a contains all the elements in s.
+func contains(a map[string]bool, s ...string) bool {
+	var r bool
+	for _, e := range s {
+		r = a[e]
+	}
+	return r
```
davidkaya: This implementation is not correct, it will return whether the last string from `s` is in the set, not all of them.

davecheney: Thanks for spotting my bug.

davecheney: Please see e1c1f75

PR #180, opened 2015-06-02
cmd/gb-vendor/vendor: add test for vendor.ImportPaths

davecheney:


cmd/gb-vendor/vendor/_testdata/src/github.com/foo/bar/main.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/quux/flobble"
+	// "bitbucket.org/fwoop/ftang" // commented out, this is deliberate
```
davidkaya: We could also add something like `import flobble "github.com/quux/flobble"` or `import _ "github.com/quux/flobble"`.

davecheney: Go for it, add more test fixtures, including a _testdata somewhere in there
and adjust the expected result.

cmd/gb-vendor/vendor/imports_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+func TestParseImports(t *testing.T) {
+	root := filepath.Join(getwd(t), "_testdata")
+
+	got, err := ParseImports(root)
+	if err != nil {
+		t.Fatalf("ParseImports(%q): %v", root, err)
+	}
+
+	want := set("fmt", "github.com/quux/flobble")
```
davidkaya: This won't work, because `ParseImports` does not return packages from stdlib.

davecheney: Clearly it does, the test passes.

davecheney: Merged latest fixes to stringset.contains, test still passes

davidkaya: Isn't that weird or I got something wrong?

davecheney: The bug is the s.Path.Value returns quoted versions, we don't strip the quotes unless it does not match the standard library, which it never will, because quotes.

davecheney:
Please fix up `ParseImports` and add the extra tests we discussed.

PR #183, opened 2015-06-03
cmd/gb-vendor: add support for vanity imports

### LINKED ISSUE #178: cmd/gb-vendor: fetch/update support vanity imports
This is a tracking issue to add support for dynamic imports (commonly known as vanity imports).

davecheney:
Fixes #177 
Fixes #178 

This PR adds initial support to `gb vendor` for handling so called 'vanity' import paths.

Included are some regrettable tests which require network access, this is probably unavoidable without mocking out the http layer, so for the moment we'll live with it.

```
 % gb vendor fetch golang.org/x/crypto/pbkdf2
Cloning into '/tmp/gb-vendor-402709523'...
remote: Total 1922 (delta 1207), reused 1922 (delta 1207)
Receiving objects: 100% (1922/1922), 1.42 MiB | 403.00 KiB/s, done.
Resolving deltas: 100% (1207/1207), done.
Checking connectivity... done.
copyfile(dst: /home/dfc/devel/juju/vendor/src/golang.org/x/crypto/pbkdf2/pbkdf2.go, src: /tmp/gb-vendor-402709523/pbkdf2/pbkdf2.go)
copyfile(dst: /home/dfc/devel/juju/vendor/src/golang.org/x/crypto/pbkdf2/pbkdf2_test.go, src: /tmp/gb-vendor-402709523/pbkdf2/pbkdf2_test.go)
```

### DIFF
--- a/cmd/gb-vendor/vendor/imports.go
+++ b/cmd/gb-vendor/vendor/imports.go
@@ -1,8 +1,11 @@
 package vendor
 
 import (
+	"fmt"
 	"go/parser"
 	"go/token"
+	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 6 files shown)


cmd/gb-vendor/vendor/imports.go
```diff
@@ -40,3 +43,28 @@ func ParseImports(root string) (map[string]bool, error) {
...
+	if err != nil {
+		return "", "", "", err
+	}
+	defer rc.Close()
+
+	meta, err := parseMetaGoImports(rc)
+	if err != nil || len(meta) < 1 {
+		return "", "", "", err
+	}
+	return meta[0].Prefix, meta[0].VCS, meta[0].RepoRoot, nil
```
davidkaya: Subjective opinion: Wouldn't it be "nicer" if we return struct with Prefix, VCS and RepoRoot attributes instead of 3 return values + error?

davecheney: Yeah. That's a fair comment, this code is pretty rough and ready. The
metaImport struct comes from discover.go, which I imported from
x/toosl/go/vcs, I dunno if I want to keep it, by parsing that xml is messy.

I'm also uncomfortable with passing around a Repository interface value
_and_ an extra bit of path, but I don't think I understand the problem well
enough to know the right data structure -- but hey, at least I got rid of
the MultiRepo type.

No matter how ugly, I want to get this into the hands of testers as soon as
possible, then refactor from there.

davidkaya: Ok! It was my only comment that I've found, lgtm, you can merge it.

davecheney: Yeah. Sorry to pull rank. I know this code sucks, but it works (at least in
the success case) so I'd like to unblock all the people waiting on
golang.org/x/ repos.

PR #191, opened 2015-06-07
cmd/gb-vendor: merge wc.CheckoutBranch/CheckoutRevision into RemoteRepo.Checkout

davecheney:


cmd/gb-vendor/vendor/repo.go
```diff
@@ -137,13 +134,22 @@ func (g *gitrepo) URL() string {
...
 	}
 
-	if err := runOut(os.Stderr, "git", "clone", g.url, dir); err != nil {
+	args := []string{
+		"clone",
+		g.url,
+		dir,
+	}
+	if branch != "" {
+		args = append(args, "--branch", branch)
```
davidkaya: I would also add `--single-branch`, that will clone only that specified branch. We don't need other branches and that will reduce size of the downloaded(cloned) repo.

cmd/gb-vendor/vendor/repo.go
```diff
@@ -212,13 +208,21 @@ type hgrepo struct {
...
+		h.url,
+		dir,
+	}
 
-	if err := runOut(os.Stderr, "hg", "clone", h.url, dir); err != nil {
+	if branch != "" {
+		args = append(args, "--branch", branch)
+	}
+	if err := runOut(os.Stderr, "hg", args...); err != nil {
 		return nil, err
```
davidkaya: Suggestion: Maybe `os.RemoveAll(dir)` when the cloning fails to cleanup the temp dir.

cmd/gb-vendor/vendor/repo.go
davecheney: Done

>  Also you are not using revision anywhere.

how should I use revision now ? maybe to a checkout rev ?

davecheney:
@davidkaya please take another look

cmd/gb-vendor/vendor/repo.go
davidkaya: Well honestly I would remove the support for revisions completely. Anyone who checks out revisions in projects, either does not know how to use vcs or wants to do something completely silly which can be done better (of course there could be some cases where this could be needed, but I have never seen such).

That would also remove the problem with `Detached HEAD`. 

Nevertheless, you can checkout either branch or commit. You can't do both at once. User can't specify branch and revision at once, there must be a check if user didn't do that. 

If you want to checkout commit (revision) you can do something like this:
```bash
git clone $repo
git reset --hard $revision
```

davecheney: >
> Nevertheless, you can checkout either branch or commit. You can't do both
> at once. User can't specify branch and revision at once, there must be a
> check if user didn't do that.
>
I think you can do both on mercurial, commits are related to branches, ie
some hash may only be valid on a specific branch.


>  If you want to checkout commit (revision) you can do something like this:
>
> git clone $repo
> git reset --hard $revision
>
> You know a lot more about this than I do, do you want to do a change when
#191 lands ?

davidkaya: Yes I can.

PR #194, opened 2015-06-07
Support older versions of git (gb-vendor)

42wim:
Git (1.8.3.1) on centos/rhel 7 doesn't support -C
Specify the working directory when executing git instead of using -C

### DIFF
--- a/cmd/gb-vendor/vendor/repo.go
+++ b/cmd/gb-vendor/vendor/repo.go
@@ -156,7 +156,7 @@ func (g *gitrepo) Checkout(branch, revision string) (WorkingCopy, error) {
 	}
 
 	if revision != "" {
-		if err := runOut(os.Stderr, "git", "-C", dir, "checkout", revision); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/gb-vendor/vendor/repo.go
```diff
@@ -357,3 +357,17 @@ func runOut(w io.Writer, c string, args ...string) error {
 	cmd.Stderr = os.Stderr
 	return cmd.Run()
 }
+
+func runPath(path string, c string, args ...string) ([]byte, error) {
+	var buf bytes.Buffer
+	err := runOut(&buf, c, args...)
```
davecheney: did you mean to call runOutPath here

davecheney:
Thanks for testing this. CentOS is why we cannot have nice things.

42wim:
No, thank you for making this :) I'm also surprised centos 7 has such an old git version.
Rebased the PR with the runOutPath fixed.

PR #195, opened 2015-06-08
Clean up temporary files on build error

42wim:
When doing a failed build the temporary directories don't get removed.


42wim:
As far as I can see, it can fail, but it won't give an error to the user if it fails. It's just unhandled. (or am I overlooking something?)
Do you really want error handling there ?

davidkaya:
Well, if I was the user, I would want to know, that the clean up failed. Now the user will just know, that build failed and he won't know that the clean up also failed.

Nevertheless, I would let Dave decide what is right.

davecheney:
Nah, this is the correct choice. Thanks @42wim

PR #205, opened 2015-06-10
style(vendor): fix go vet format error

krhubert:
add "path" argument to assertNotExists and assertExists functions


davecheney:
Thanks

Issue #136, opened 2015-05-26
gb-vendor: fetch / update: vendor specific revision of a package 


davecheney:
I was thinking that it would be nice if gb-vendor had additional arguments, where user would be able to specify commit/branch of a package he or she wants to vendor.

Something like gb-vendor -c 21dc14a51f github.com/user/sample or gb-vendor -b branch_name github.com/user/sample

davecheney:
flags should be
`gb vendor $REPO --branch $branch --revision $revision`

davidkaya:
@davecheney 
`--revision` this is commit sha?

davecheney:
Yes. I feel strongly that go packages do not have versions (at the moment).

pires:
I would vote for `tags` instead of `branches`.

PSG-Luna:
I think having both makes sense

rnapier:
Is a special argument needed to handle tags? For git, most "treeish" things are equivalent syntax, but that's not universal.

I assume still the actual SHA should be recorded. Will Manifest need to record how that SHA was selected (branch, SHA, tag)?

If you don't pass a branch, how does update work? Will it always update to master even if the original SHA was on a branch?

drewwells:
Both would be handy, but sha is very nice for now

davidkaya:
I have one suggestion. `gb vendor update $repo` should not have `--branch` or `--revision` tag, I think it should only require `$repo`, because the branch/revision are specified in the manifest. I think that `gb vendor update $repo` should only pull latest changes for the branch specified in the manifest file (revision is irrelevent, because after updating, there will be new revision).

On the other hand, if a user wants to switch branch/revision of the package, I would suggest new command `gb vendor rebase --branch $branch $repo` or `gb vendor rebase --revision $rev $repo`.

davecheney:
Yup, that makes sense.

Instead of rebase, I'd prefer people do remove and then fetch.

Gb vendor is just a convenience wrapper around an action they could take
manually, so I'm not interested in addressing extreme cases.

davidkaya:
Implemented in #132

PR #206, opened 2015-06-10
gb-vendor: Adds support for the -branch flag to the update command

dcarney:
First off, thanks for all the hard work on gb!  I'm enthusiastically converting projects over to use it.

This PR adds support for the `-branch` flag to `gb vendor update` command. The flag works just as it does when using `gb vendor fetch`.

This allows you to take a currently vendored dependency and update it to the HEAD revision of a branch other than `master`.

Currently, the only way to achieve this functionality is with:
`$ gb vendor delete github.com/foo/bar`
`$ gb vendor update -branch=bugfix github.com/foo/bar`

which seems a little unnecessary.

### DIFF
--- a/cmd/gb-vendor/update.go
+++ b/cmd/gb-vendor/update.go
@@ -20,17 +20,22 @@ var (
 
 func addUpdateFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&updateAll, "all", false, "update all dependencies")
+	fs.StringVar(&branch, "branch", "master", "branch of the package")
 }
 
 var cmdUpdate = &cmd.Command{
 	Name:      "update",
-	UsageLine: "update [-all] import",
+	UsageLine: "update [-branch branch] [-all] import",
 	Short:     "update a local dependency",
-	Long: `gb vendor update will replaces the source with the latest available from the head of the master branch.
+	Long: `gb vendor update will replace the source with the latest available from the head of the default upsteam branch, or the branch specified with -branch.
 
 Flags:
 	-all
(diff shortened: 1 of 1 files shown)


davecheney:
Assigning to @davidkaya as he has some strong feelings on the matter.

davecheney:
@dcarney can you please update alldocs.go, inside cmd/gb-vendor, do `go install`, then `go generate`, which should rebuild `alldocs.go`.

davidkaya:
We agreed in #136 that we won't add `branch` to `gb update` and it should by default update to the HEAD of the branch saved in the Manifest file. If a user wants to switch to other branch then he should delete/fetch the repository. Because updating to other branch is no longer updating but it is changing version.

However you found a bug, which probably occurred when Dave refactored the `repo` package. Because it should not be `repo.Checkout("", "")`, it should checkout the branch specified in the manifest file.

In the beginning of the cycle, there should be `branch := d.Branch` and that branch should be checked out.

davecheney:
@dcarney I defer to @davidkaya on this one, I think his explanation that if you want to switch branches you should delete/fetch; as this is effectively what update is doing.

Thanks for finding the bug, I'd appreciate it if one of you sent a PR to fix it.

I'll consider this issue closed when fetch/update is documented to explain the workflow we expect.

Thanks

Dave

davecheney:
@dcarney #211 fixes the bug i introduced. If you're happy with our explanation of rejecting -branch on update then I can close this with an update to the documentation.

How does that sound ?

dcarney:
Sounds good. Thanks for the explanation.

PR #211, opened 2015-06-11
Fixed and refactored update

davidkaya:
Fixed bug that was found thanks to @dcarney's PR (#206) and refactored code.

### DIFF
--- a/cmd/gb-vendor/update.go
+++ b/cmd/gb-vendor/update.go
@@ -59,25 +59,23 @@ Flags:
 		}
 
 		for _, d := range dependencies {
-			url := d.Repository
-			p := d.Importpath
-
 			err = m.RemoveDependency(d)
 			if err != nil {
 				return fmt.Errorf("dependency could not be deleted from manifest: %v", err)
 			}
 
-			if err := os.RemoveAll(filepath.Join(ctx.Projectdir(), "vendor", "src", filepath.FromSlash(p))); err != nil {
+			if err := os.RemoveAll(filepath.Join(ctx.Projectdir(), "vendor", "src", filepath.FromSlash(d.Importpath))); err != nil {
 				// TODO(dfc) need to apply vendor.cleanpath here to remove indermediate directories.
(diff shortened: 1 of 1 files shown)


cmd/gb-vendor/update.go
```diff
@@ -59,45 +59,38 @@ Flags:
...
-			url := d.Repository
-			p := d.Importpath
-
 			err = m.RemoveDependency(d)
 			if err != nil {
 				return fmt.Errorf("dependency could not be deleted from manifest: %v", err)
 			}
 
-			if err := os.RemoveAll(filepath.Join(ctx.Projectdir(), "vendor", "src", filepath.FromSlash(p))); err != nil {
+			if err := os.RemoveAll(filepath.Join(ctx.Projectdir(), "vendor", "src", filepath.FromSlash(d.Importpath))); err != nil {
```
davidkaya: @davecheney What do you think about this long line? Wouldn't it be more readable if it was split into multiple lines?

davecheney: I'm fine with long lines, but if you want to make it shorter pull the filepath.Join part out into another line.

cmd/gb-vendor/update.go
```diff
@@ -59,45 +59,38 @@ Flags:
...
-			branch, err := wc.Branch()
+			rev, err := wc.Revision()
 			if err != nil {
 				return err
 			}
 
 			dep := vendor.Dependency{
-				Importpath: p,
-				Repository: url,
+				Importpath: d.Importpath,
```
davecheney: I think we should be using the values from the fresh checkout here, rather than the original values from d. 

I think this because at best they are identical, in which case there is no semantic difference, but at worse, if they have changed, and they could, because DeduceRepo will get smarter over time, if we use the old values from d, then we won't be recording the updated state in the manifest.

davecheney:
LGTM, with my comments about like 97+ addressed. Thanks for fixing the bug.

davidkaya:
Included suggested changes.

davecheney:
LGTM. Ship it.

PR #214, opened 2015-06-11
gb-vendor: add support for fetching a tag

### LINKED ISSUE #209: cmd/gb-vendor: add support for fetching from a tag
@davidkaya what do you say to adding support for gb vendor fetching from a tag, not a revision ?

Updates docker/docker#13756

davecheney:
Fixes #209

This PR adds support for fetching from a tag for all repos, however
tag is ignored by all but git. Hopefully someone with hg or bzr
skills can improve this.

This PR also adds a check to ensure that if a repo, again git, is fetched
from a tag or revision, that update will not let the caller update, as
there is no path from one revision to the next if you are in a detached
state.

The error message is very bad, but at least we don't promise something that
isn't possible.
```
% gb vendor fetch -tag r2015.06.03 gopkg.in/mgo.v2
% gb vendor update gopkg.in/mgo.v2
FATAL command "update" failed: cannot checkout HEAD
```

### DIFF
--- a/cmd/gb-vendor/fetch.go
+++ b/cmd/gb-vendor/fetch.go
@@ -13,28 +13,33 @@ import (
 var (
 	// gb vendor fetch command flags
 
-	// branch
 	branch string
 
 	// revision (commit)
 	revision string
+
+	tag string
 )
 
 func addFetchFlags(fs *flag.FlagSet) {
 	fs.StringVar(&branch, "branch", "master", "branch of the package")
 	fs.StringVar(&revision, "revision", "", "revision of the package")
+	fs.StringVar(&tag, "tag", "", "tag of the package")
 }
 
 var cmdFetch = &cmd.Command{
 	Name:      "fetch",
-	UsageLine: "fetch [-branch branch] [-revision rev] importpath",
+	UsageLine: "fetch [-branch branch] [-revision rev|-tag tag] importpath",
 	Short:     "fetch a remote dependency",
 	Long: `fetch vendors the upstream import path.
 
 Flags:
 	-branch branch
 		fetch from the name branch. If not supplied the default upstream
-		branch will be used
+		branch will be used.
+	-tag tag
+		fetch the specified tag. If not supplie the default upstream 
+		banch will be used.
 	-revision rev
 		fetch the specific revision from the branch (if supplied). If no
 		revision supplied, the latest available will be supplied.
@@ -60,7 +65,7 @@ Flags:
 			return fmt.Errorf("%s is already vendored", path)
 		}
 
-		wc, err := repo.Checkout(branch, revision)
+		wc, err := repo.Checkout(branch, tag, revision)
 		if err != nil {
 			return err
 		}
--- a/cmd/gb-vendor/fetch.go
+++ b/cmd/gb-vendor/fetch.go
@@ -29,7 +29,7 @@ func addFetchFlags(fs *flag.FlagSet) {
 
 var cmdFetch = &cmd.Command{
 	Name:      "fetch",
-	UsageLine: "fetch [-branch branch] [-revision rev|-tag tag] importpath",
+	UsageLine: "fetch [-branch branch | -revision rev | -tag tag] importpath",
 	Short:     "fetch a remote dependency",
 	Long: `fetch vendors the upstream import path.
 
@@ -39,7 +39,7 @@ Flags:
 		branch will be used.
 	-tag tag
 		fetch the specified tag. If not supplie the default upstream 
-		banch will be used.
+		branch will be used.
 	-revision rev
 		fetch the specific revision from the branch (if supplied). If no
 		revision supplied, the latest available will be supplied.
--- a/cmd/gb-vendor/fetch.go
+++ b/cmd/gb-vendor/fetch.go
@@ -22,7 +22,7 @@ var (
 )
 
 func addFetchFlags(fs *flag.FlagSet) {
-	fs.StringVar(&branch, "branch", "master", "branch of the package")
+	fs.StringVar(&branch, "branch", "", "branch of the package")
 	fs.StringVar(&revision, "revision", "", "revision of the package")
 	fs.StringVar(&tag, "tag", "", "tag of the package")
 }
--- a/cmd/gb-vendor/vendor/repo.go
+++ b/cmd/gb-vendor/vendor/repo.go
@@ -153,20 +153,16 @@ func (g *gitrepo) URL() string {
 	return g.url
 }
 
-// Checkout fetchs the remote branch, and optionaly updates to the supplied
-// tag or revision. If both are supplied, an error is returned. If a branch
-// and a tag is supplied, then an error is return. If the branch is blank,
+// Checkout fetchs the remote branch, tag, or revision. If more than one is
+// supplied, an error is returned. If the branch is blank,
 // then the default remote branch will be used. If the branch is "HEAD", an
 // error will be returned.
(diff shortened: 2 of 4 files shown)


cmd/gb-vendor/vendor/repo.go
```diff
@@ -154,7 +153,17 @@ func (g *gitrepo) URL() string {
...
 }
 
-func (g *gitrepo) Checkout(branch, revision string) (WorkingCopy, error) {
+// Checkout fetchs the remote branch, and optionaly updates to the supplied
+// tag or revision. If both are supplied, an error is returned. If the branch
+// is blank, then the default remote branch will be used. If the branch is
+// "HEAD", an error will be returned.
+func (g *gitrepo) Checkout(branch, tag, revision string) (WorkingCopy, error) {
+	if branch == "HEAD" {
+		return nil, fmt.Errorf("cannot checkout HEAD")
```
davecheney: given this can only happen when `gb vendor update $SOMETHING_TAGGED` I could change this to

"cannot update dependency that has been previously fetched with -tag or -revision"

cmd/gb-vendor/vendor/repo.go
```diff
@@ -176,8 +185,8 @@ func (g *gitrepo) Checkout(branch, revision string) (WorkingCopy, error) {
 		return nil, err
 	}
 
-	if revision != "" {
-		if err := runOutPath(os.Stderr, dir, "git", "checkout", revision); err != nil {
+	if revision != "" || tag != "" {
+		if err := runOutPath(os.Stderr, dir, "git", "checkout", "-q", oneOf(revision, tag)); err != nil {
```
davidkaya: If user specifies branch and tag/revision which is not in that branch it will fail. `--single-branch` clones only that branch.

Nevertheless specifying branch and tag at once should not work at all.

davecheney:
Done, the new error message is a little verbose, but gets the job done
```
% gb vendor update gopkg.in/mgo.v2
FATAL command "update" failed: cannot update "https://gopkg.in/mgo.v2" as it has been previously fetched with -tag or -revision. Please use gb vendor delete then fetch again.
```

cmd/gb-vendor/fetch.go
```diff
@@ -13,28 +13,33 @@ import (
...
 	Short:     "fetch a remote dependency",
 	Long: `fetch vendors the upstream import path.
 
 Flags:
 	-branch branch
 		fetch from the name branch. If not supplied the default upstream
-		branch will be used
+		branch will be used.
+	-tag tag
+		fetch the specified tag. If not supplie the default upstream
```
davidkaya: supplie -> supplied

cmd/gb-vendor/vendor/repo.go
```diff
@@ -154,7 +153,17 @@ func (g *gitrepo) URL() string {
...
-func (g *gitrepo) Checkout(branch, revision string) (WorkingCopy, error) {
+// Checkout fetchs the remote branch, and optionaly updates to the supplied
+// tag or revision. If both are supplied, an error is returned. If the branch
+// is blank, then the default remote branch will be used. If the branch is
+// "HEAD", an error will be returned.
+func (g *gitrepo) Checkout(branch, tag, revision string) (WorkingCopy, error) {
+	if branch == "HEAD" {
+		return nil, fmt.Errorf("cannot update %q as it has been previously fetched with -tag or -revision. Please use gb vendor delete then fetch again.", g.url)
+	}
+	if !atMostOne(tag, revision) {
```
davidkaya: branch+tag or branch+revision also cannot be supplied

cmd/gb-vendor/vendor/repo.go
davecheney: > Nevertheless specifying branch and tag at once should not work at all. 

added a check

> If user specifies branch and tag/revision which is not in that branch it will fail. --single-branch clones only that branch.

I don't know if we can do much to help there, I really don't want to get into the business of wrapping the whole git checkout process -- those flags are really only if you want exactly that specific version of a commit.

davecheney:
PTAL. I had to change the default value of -branch to not default to master.

davidkaya:
Now it will fail if you do `gb fetch github.com/repo/r` because it will call `git checkout --branch github.com/repo/r --single-branch` (there should be some branch specified).

davecheney:
I think this will be ok, because we checl
```
        if branch != "" {
                args = append(args, "--branch", branch)
        }
```

davidkaya:
Oh you are right, sorry.

davecheney:
I've checked all the permutations and I think this is now good to land

fetch -branch && update => update latest on branch
fetch && update => update to latest on default branch
fetch -tag && update => fail, detached head
fetch -tag -branch => fail, only one of branch, tag, revision

davidkaya:
aaaa wait. It will still not work. If you want to checkout tag, you have to put "tags/" in front of the name of the tag.

`git checkout tags/<tag>`

or you can just clone it as branch
`git clone --branch <tag> $repo --single-branch`

davecheney:
Seems to work anyway
```
lucky(~/devel/juju) % gb vendor fetch -tag r2014.10.12 gopkg.in/mgo.v2


lucky(~/devel/juju) % gb vendor list | grep gopkg.in/mgo.v2
gopkg.in/mgo.v2/bson    https://gopkg.in/mgo.v2/bson    master
 3569c88678d88179dcbd68d02ab081cbca3cd4d0
gopkg.in/mgo.v2 https://gopkg.in/mgo.v2 HEAD
 231ce7e0549b5ce60d35ed805bff505cab7098e4
```

The hash matchs, https://github.com/go-mgo/mgo/tree/r2014.10.12

davidkaya:
Well that is weird. Ok, if it works its fine.

tianon:
I've tested this out and it works pretty well, but I'm curious why the tag that was fetched isn't stored in the manifest?

When I mentioned the feature originally, the main point of using an explicit tag in Docker's case (besides the obvious reason of using official upstream releases) is to make it clear to packagers which versions of our dependencies we require.  With this as-is, it'd be a non-trivial amount of work for a packager to look at the following manifest and realize that `github.com/coreos/go-systemd` is at `v2`, not just some random ref:

```json
{
	"version": 0,
	"dependencies": [
		{
			"importpath": "github.com/coreos/go-systemd",
			"repository": "https://github.com/coreos/go-systemd",
			"revision": "f743bc15d6bddd23662280b4ad20f7c874cdd5ad",
			"branch": "HEAD"
		}
	]
}
```

davidkaya:
That was the first thing that I have pointed out to @davecheney that without the `tag` in Manifest, users won't be able to (at least easily) tell which version of a package they use.

I think he will reconsider it. Also it will be trivial to add such feature.

davecheney:
@davidkaya is correct. I punted on adding a tag field to the manifest as this would mean bumping the manifest version, and that would mean writing and testing upgrade code. 

When I have to bump the manifest version, i'll add a tag field.

@tianon you wrote
> When I mentioned the feature originally, the main point of using an explicit tag in Docker's case (besides the obvious reason of using official upstream releases) is to make it clear to packagers which versions of our dependencies we require.

If (in some parallel universe) Docker was built as a gb project, why would packagers care which revisions (go packages don't have versions) of a package were imported at ? I'm guessing they either want to package those dependencies externally (puke), or they want to sprinkle security by somehow tracking the version (which Go packages don't have) of potentially vulnerable packages.

tianon:
That puke reaction is exactly what's happening today. :wink:

PR #218, opened 2015-06-12
Use pkg/exec for exec handling

davecheney:
gb had at least three copies of helper functions for handling running commands, running commands and capturing output, running commands and changing the working directory, etc. The combinational explosion of these permutations was annoying, not to mention the duplication.

Refactor this into a new package, `github.com/pkg/exec` which uses a function options pattern on top of the standard `os/exec` package.

This will also be the first non standard library package that gb relies on, and until we're self hosted, we'll just have to live with it. Depending on how painful this is, self hosting will come either sooner or later.


davidkaya:
Lgtm. Nice refactoring, it looks a lot better now.

davecheney:
Thanks. I was incensed by the requirement to add another pair of run* functions because the broke ass version of git on Centos7 didn't support -C, the thought of having to continue to add permutations for different variants of working directory, output capture, environment munging, etc pushed me over the edge.

davidkaya:
Tbh. I admire people who use Centos ^^

davecheney:
I'm going to fork pkg/exec into gb for now. I don't want to tackle the dependency issue at the moment

PR #223, opened 2015-06-16
Support for optional fetching via insecure paths

### LINKED ISSUE #207: gb vendor fetch cannot retrieve http hosted (non-secure) packages
I have several general purpose packages in our internal VCS. These packages are used in multiple projects each of which should vendor them for reproducible project builds. But our internal VCS does not have https enabled.

Any chance `gb vendor fetch` could try http if https is not available for a repo?

davecheney:
Fixes #207

Adds support for optional fetching via insecure paths.

`gb fetch` / `update` have a new flag, `--precaire` which enables discovery of insecure repos. It defaults to off.

### TODO
No support for insecure bzr, I'll wait for someone to request it.

### For discussion
Currently `git://` scheme is considered insecure as, while it has no support for pushing, there is no transport security so could be used in a man-in-the-middle attempt.

Who cares enough to debate this ?

### DIFF
--- a/cmd/gb-vendor/alldocs.go
+++ b/cmd/gb-vendor/alldocs.go
@@ -42,6 +42,8 @@ Flags:
 	-revision rev
 		fetch the specific revision from the branch (if supplied). If no
 		revision supplied, the latest available will be supplied.
+	-precaire
+		allow the use of insecure protocols.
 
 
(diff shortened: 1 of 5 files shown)


davecheney:
@ChrisHines @jpoehls this is ready for review.

Issue #228, opened 2015-06-18
vendor update does not leave project in a consistent state after errors


ChrisHines:
If gb vendor encounters an error before completing, it may have updated `vendor/src/...` but the manifest file is left unchanged.

This issue comes up repeatedly on Windows where deleting a temporary git clone always fails (see  #200).

A workaround is:

1. `gb vendor delete <import path>`
2. `gb vendor fetch <import path>`

davecheney:
I agree we can do better here, by moving the removal of the old dep down later, and writing the manifest before removing the temporary copy.

There is still a window where failures can happen, and some of that responsibility for covering that will have to be covered by project owners to use source control studiously.

PR #229, opened 2015-06-19
Write manifest after each update.

ChrisHines:
Fix for #228.

### DIFF
--- a/cmd/gb-vendor/update.go
+++ b/cmd/gb-vendor/update.go
@@ -119,12 +119,16 @@ Flags:
 				return err
 			}
 
+			if err := vendor.WriteManifest(manifestFile(ctx), m); err != nil {
+				return err
+			}
+
 			if err := wc.Destroy(); err != nil {
 				return err
 			}
 		}
 
-		return vendor.WriteManifest(manifestFile(ctx), m)
+		return nil
 	},
 	AddFlags: addUpdateFlags,
 }
--- a/cmd/gb-vendor/update.go
+++ b/cmd/gb-vendor/update.go
@@ -70,6 +70,11 @@ Flags:
 		}
 
 		for _, d := range dependencies {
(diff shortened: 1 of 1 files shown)


cmd/gb-vendor/update.go
```diff
@@ -119,12 +119,16 @@ Flags:
 				return err
 			}
```
davecheney: github won't let me comment on line 73, but please move line 73 down to around 117, the sequence should be

1. fetch dependency into tempdir
2. remove old copy
3. copypath the new copy
4. write back manifest
5. wc.Destroy

davecheney:
Thanks for this. Please review the comments i've made and I can merge this for you.

cmd/gb-vendor/update.go
ChrisHines: Ah yes, good catch. One moment.

cmd/gb-vendor/update.go
```diff
@@ -100,6 +90,11 @@ Flags:
...
 
+			err = m.RemoveDependency(d)
```
davecheney: I am very sorry, I told you the wrong thing, m.RemoveDependency(d) checks that the dependency is present in the manifest. It should be above the call to DeduceRepo.

cmd/gb-vendor/update.go
```diff
@@ -112,6 +107,11 @@ Flags:
 				return err
 			}
 
+			if err := os.RemoveAll(filepath.Join(ctx.Projectdir(), "vendor", "src", filepath.FromSlash(d.Importpath))); err != nil {
```
davecheney: this is fine, please don't move this

davecheney:
Sorry, one final tweek. I'm sorry I told you the wrong thing earlier.

ChrisHines:
No worries.

PR #230, opened 2015-06-19
Windows: Make sure all files are writable before deleting a git clone.

### LINKED ISSUE #200: fetch failed on windows
when trying to fetch package github.com/labstack/echo i've got error.
OS Windows 7 x64
git version 1.9.4.msysgit.0

```
E:\go_dev\vd-crm-prj
λ gb vendor fetch github.com/labstack/echo
FATAL command "fetch" failed: remove E:\go_dev\APPDATA\tmp\gb-vendor-240678526\.git\objects\pack\pack-3461fa99fcbbe64c91c25f698a8d530602fb28d2.idx: Access is denied.
```

help to fix this problem

ChrisHines:
Fixes #200.

### DIFF
--- a/cmd/gb-vendor/delete.go
+++ b/cmd/gb-vendor/delete.go
@@ -3,7 +3,6 @@ package main
 import (
 	"flag"
 	"fmt"
-	"os"
 	"path/filepath"
 
 	"github.com/constabulary/gb"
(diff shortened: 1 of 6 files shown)


davecheney:
Thanks for working on this. I don't think this fix is correct, this logic shouldn't require creating a fake working copy just to use it's Destroy method.

Can you please move your additions into a RemoveAll method inside `cmd/gb-vendor`, then replace all the instances of `os.RemoveAll` with this version.

Thanks

ChrisHines:
Sure, although it should probably wait until my other PR is merged to avoid conflicts.

PR #231, opened 2015-06-19
cmd/gb: add experimental depset command and add recursive fetch ability

davecheney:
depset is an experimental command to explore analysing dependency sets.
It produces output like this, http://paste.ubuntu.com/11739677/

At the moment it has worse than quadratic behaviour, making it unsuitable for anything serious.

The goal of depset is to develop a set of functions in the `cmd` package that can be used by tools like `gb-vendor` to analyse dependencies as a set rather than as individual imports.
Dependency sets are groups of packages that are related by being children of a single source root. A canonical example of a source root is $GOPATH/src, another is $PROJECT/src. However we can be smarter and create sets for each vendored import tree, which in turn allows us to determine which dependencies are no longer referenced with more accuracy than the current gb vendor purge command.

Another function of dep sets is to determine missing packages, the output of which could be fed back into gb vendor fetch to drive recursive depdendency fetching.

/cc @davidkaya

### DIFF
--- a/cmd/depset.go
+++ b/cmd/depset.go
@@ -1,0 +1,144 @@
+package cmd
+
+import (
+	"fmt"
+	"go/build"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func keys(m map[string]bool) []string {
+	var v []string
+	for k := range m {
+		v = append(v, k)
+	}
+	return v
+}
+
+// Pkg describes a Go package.
+type Pkg struct {
+	*Depset
+	*build.Package
+}
+
+// Depset describes a set of related Go packages.
+type Depset struct {
+	Root   string
+	Prefix string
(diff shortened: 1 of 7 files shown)


davidkaya:
As soon as I get out of work I will look into it.

davecheney:
Made the search non quadratic, now works for a much larger dataset

http://paste.ubuntu.com/11739713/

davecheney:
No hurry. This isn't ready for review, only for your info. This is the
surprise I talked about earlier.

davidkaya:
@davecheney Is it just me or the Wercker is weird? I see it as "Waiting to hear about 7332c04" for 3 days already.

davecheney:
No. I do not think it is you, I think wercker may be unwell.

PR #235, opened 2015-06-23
Fix typos and missed option

agonzalezro:
- Fixed typos in several place were the word dependency was misspelled.
- Add `-precaire` option to the places of the documentation where it was
  missing.


agonzalezro:
That was quick! Thanks.

davecheney:
No, thank you for spotting these. It is very much appreciated.

Issue #31, opened 2015-05-02
Cross compilation support


davecheney:
Related to #9 and #26 cross compilation with gb should always just work.

Right now it does not and there are a few parts to solving this

- The first is is #28 because not only does the toolchain change between go 1.4 and go 1.5, but so does the procedure for creating a cross compilation capable toolchain.

- The second is related to the first. Specifically in Go 1.5 we removed the requirement to do the `./make.bash --no-clean` dance by letting the `go` tool rebuild all the way back to `runtime` if necessary.

For `gb`, I never intended it to be a tool that would compile the Go standard lib itself, so any package which is considered to be the standard library is assumed to be up to date and never recompiled, with Go 1.5, this won't be sufficient.

Actions needed are at least
- icky interaction between `gb.gcToolchain` and the package selection logic to determine if it needs to rebuild a std lib package
- decide on #28 
- enhancements to gb to build, but _not_ install into the toolchains' $GOROOT, packages.

davecheney:
#37 is a prerequisite of this issue.

davecheney:
Cross compilation is explicitly disabled in d5a861b6f83f06511871144e1759553d4e67e877 and now you get an error telling you why
```
% env GOOS=darwin gb build ./...
FATAL unable to construct context: cross compilation from host linux/amd64 to target darwin/amd64 not supported. See issue #31
```

PR #243, opened 2015-06-28
Add Cross compilation support

mopemope:
Cross compilation support #31
Usage simple

```
$ gb build
...
$ ls bin
csv

```

```
$ GOOS=windows GOARCH=amd64 gb build
...
$ ls bin
csv-windows-amd64.exe
```

### DIFF
--- a/gc14.go
+++ b/gc14.go
@@ -8,15 +8,25 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 )
 
 func GcToolchain(opts ...func(*gcoption)) func(c *Context) error {
+	goos := runtime.GOOS
+	goarch := runtime.GOARCH
+
 	defaults := []func(*gcoption){
 		func(opt *gcoption) {
 			opt.goos = runtime.GOOS
+			if v := os.Getenv("GOOS"); v != "" {
+				opt.goos = v
+			}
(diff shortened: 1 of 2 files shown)


davecheney:
Thank you for this pull request. I don't want to merge this at this time for the following reasons

1. cross compilation requires a working compiler and standard library, this differs between Go 1.4 and Go 1.5 and the requirements for getting a good cross compile need to be documented. At the moment if these conditions are not met, they will create a large support load.

2. For Go 1.5, gb should be able to build the standard library for the target arch if it is missing, but this is not implemented yet, #32. This should be completed first.

3. It is not clear to me how this change will interact with conditional compilation and I cannot test this without resolving the issues above. 

4. This change causes the GOOS and GOARCH seen by gctoolchain to differ from the values recorded in gc.Context. The latter is used for conditional compilation to determine the correct path the cached package archive. As a first step, setting GOOS and GOARCH from the env should be plumbed through http://godoc.org/github.com/constabulary/gb#Project.NewContext

I hope that we can merge this in the future. I am sorry that cross compilation has been delayed by work on gb-vendor and cgo.

davecheney:
Please see the discussion in https://github.com/constabulary/gb/pull/251#issuecomment-117388375

PR #247, opened 2015-06-29
Updated fetch.go to correct a typo in the -no-recurse flag

### LINKED ISSUE #246: Typo for no-recurse flag on gb vendor fetch
In trying to disable recursive fetching (as a workaround for #245) I attempted to utilize the -no-recurse option but I found it failing as follows:

    PS C:\git\stash\playground\cerebro> gb vendor fetch -no-recurse github.com/aws/aws-sdk-go
    flag provided but not defined: -no-recurse

So I dug into the code a bit and found the flag had a typo inside the gb-vendor command [here](https://github.com/constabulary/gb/blob/master/cmd/gb-vendor/fetch.go#L35).

Naturally if I update it to be -no-recuse=false it finds the flag and is happy. Based on the way the variable is used in code the negation in the flag is a bit counter intuitive. The flag should simply be called -recurse (in my humble opinion) to match the variable being backed by its value. So then the following -recurse=false makes sense and matches what the code will ultimately do.

wmadisonDev:
Fixes #246 

Changing -no-recurse flag to simply be called -recurse given that the default behavior is to fetch recursively unless explicitly disabled.

### DIFF
--- a/cmd/gb-vendor/fetch.go
+++ b/cmd/gb-vendor/fetch.go
@@ -23,6 +23,8 @@ var (
 	revision string
 
 	tag string
+	
+	noRecurse bool // Container variable to house the value of the noRecurse
 
 	recurse  bool // should we fetch recursively
 	insecure bool // Allow the use of insecure protocols
@@ -32,13 +34,13 @@ func addFetchFlags(fs *flag.FlagSet) {
 	fs.StringVar(&branch, "branch", "", "branch of the package")
 	fs.StringVar(&revision, "revision", "", "revision of the package")
 	fs.StringVar(&tag, "tag", "", "tag of the package")
-	fs.BoolVar(&recurse, "recurse", true, "fetch recursively")
+	fs.BoolVar(&noRecurse, "no-recurse", true, "do not fetch recursively")
 	fs.BoolVar(&insecure, "precaire", false, "allow the use of insecure protocols")
 }
 
 var cmdFetch = &cmd.Command{
 	Name:      "fetch",
-	UsageLine: "fetch [-branch branch | -revision rev | -tag tag] [-precaire] [-recurse] importpath",
+	UsageLine: "fetch [-branch branch | -revision rev | -tag tag] [-precaire] [-no-recurse] importpath",
 	Short:     "fetch a remote dependency",
 	Long: `fetch vendors the upstream import path.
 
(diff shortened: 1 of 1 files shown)


davecheney:
Thanks for fixing this, i mistook how boolean flags in go worked. 

The flag needs to be called `-no-recurse` so that the default invocation is

     gb vendor fetch $PKG

and the flag is invoked like

     gb vendor fetch -no-recurse $PKG

So the simplest way to do this is change the flag to be 

     var norecurse
    flag.BoolVar(&norecruse, "no-recurse", false, "...")

Then after flag parsing 

    recurse := !norecurse

Thanks

wmadisonDev:
Thanks for responding so quickly Dave. I'll make the suggested edits and update the PR.

cmd/gb-vendor/fetch.go
```diff
@@ -32,7 +34,7 @@ func addFetchFlags(fs *flag.FlagSet) {
 	fs.StringVar(&branch, "branch", "", "branch of the package")
 	fs.StringVar(&revision, "revision", "", "revision of the package")
 	fs.StringVar(&tag, "tag", "", "tag of the package")
-	fs.BoolVar(&recurse, "no-recuse", true, "do not fetch recursively")
+	fs.BoolVar(&noRecurse, "no-recurse", true, "do not fetch recursively")
```
davecheney: this needs to default false, we are non not recursing :)

davecheney:
Please also update the documentation, 
```
% cd gb/cmd/gb-venfor
% go install
% go generate
% go fmt 
```

cmd/gb-vendor/fetch.go
wmadisonDev: Ok you're right, just wanted to check the flagset type in package flag to make sure I wasn't crazy. So that means that to enable non-recursive fetch the user will have to do:

    -no-recurse=true

Because the default is false so if no value is specified from the cli the flag defaults to false (once I make the update), am I interpreting that correctly?

wmadisonDev:
@davecheney which documentation? The usage blurb? I thought I updated that in 6c0d2c5. If there's another spot that I missed I'll be glad to update that as well.

davecheney:
cmd/gb and cmd/gb-vendor have documentantion auto generated using go
generate in alldocs.go. Just run those steps I outline and it'll work out
fine.

wmadisonDev:
Thanks for that tip @davecheney. I ran that sequence of commands but the documentation was unaffected. I think the no-recurse flag's references in all other places were correct so the documentation was rendered correctly.

davecheney:
@wmadisonDev yes, you were right, i keep getting confused with the flags in the long doc comment and the flags package flags.

PR #252, opened 2015-07-01
gb-vendor: Order dependencies in manifest file consistently

euank:
Prior to this change, any time you ran `gb vendor update $dependency` the dependency in question would be moved to the bottom of the manifest, regardless of any other changes. This behavior also happened when adding and removing dependencies.
This led to a noisy diff where it was difficult to easily see whether a commit hash changed or if the dependency moved etc.
This change alphabetically sorts dependencies within the manifest file, thus making the git diff of revision changes, additions, and removals much cleaner looking and easier to understand.

### DIFF
--- a/cmd/gb-vendor/vendor/manifest.go
+++ b/cmd/gb-vendor/vendor/manifest.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"reflect"
+	"sort"
 )
 
 // gb-vendor manifest support
@@ -70,7 +71,7 @@ type Dependency struct {
 	// dependency was fetched from.
 	Repository string `json:"repository"`
 
-	// Revision is the revision that descibves the dependency's
+	// Revision is the revision that describes the dependency's
 	// remote revision.
 	Revision string `json:"revision"`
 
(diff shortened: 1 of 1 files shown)


davecheney:
LGTM with one change.

cmd/gb-vendor/vendor/manifest.go
```diff
@@ -137,3 +141,14 @@ func readManifest(r io.Reader) (*Manifest, error) {
 	err := d.Decode(&m)
 	return &m, err
 }
+
+// Implement sort for dependencies
+type sortableDependencies []Dependency
```
davecheney: please call this byImportpath (or ByImportpath) so you get 

     sort.Sort(byImportpath(m.Dependencies))

davecheney:
Thanks. If you can address this one little nit then i'll merge this asap

euank:
Good suggestion! Done and force-pushed over the old commit.

Thanks!

cmd/gb-vendor/vendor/manifest.go
```diff
@@ -137,3 +141,14 @@ func readManifest(r io.Reader) (*Manifest, error) {
 	err := d.Decode(&m)
 	return &m, err
 }
+
+// Implement sort for dependencies
```
davecheney: // byImport implements sort for dependencies.

davecheney:
LGTM. Thanks for your contribution.

cmd/gb-vendor/vendor/manifest.go
euank: I'm a little lazy for unexported types. Feel free to fix it or I could go back and do so if you feel strongly.

PR #258, opened 2015-07-05
Check output instead of exit status for git ls-remote

asbjxrn:
git 1.7.1 (Used by RHEL 6) does not have the --exit-code option
so gb vendor fetch fails, this changes the test to do a simple
check for the string "HEAD" in the output instead.

### DIFF
--- a/cmd/gb-vendor/vendor/repo.go
+++ b/cmd/gb-vendor/vendor/repo.go
@@ -163,7 +163,8 @@ func probeGitUrl(schemes []string, host, path string, insecure bool) (string, er
 		switch scheme {
 		case "https":
 			url := scheme + "://" + host + "/" + path
-			if _, err := run("git", "ls-remote", "--exit-code", url, "HEAD"); err == nil {
+			out, err := run("git", "ls-remote", url, "HEAD")
+			if err == nil && bytes.Contains(out, []byte("HEAD")) {
 				return url, nil
 			}
 		case "http", "git":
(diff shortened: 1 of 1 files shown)


cmd/gb-vendor/vendor/repo.go
```diff
@@ -163,7 +163,8 @@ func probeGitUrl(schemes []string, host, path string, insecure bool) (string, er
 		switch scheme {
 		case "https":
 			url := scheme + "://" + host + "/" + path
-			if _, err := run("git", "ls-remote", "--exit-code", url, "HEAD"); err == nil {
```
davecheney: i had at look at what the go tool does and it effectively does this, but does not include "--exit-code". What about just dropping the --exit-code flag ? Will that suffice ?

davecheney:
Paging @davidkaya to the operating theatre.

cmd/gb-vendor/vendor/repo.go
asbjxrn: Yes, that's probably fine. It should return error on non-existent repositories which is what's important, I guess?

davecheney: Let's go with that as it is a smaller change. We can go with your original
proposal if it proves to be insufficient.

davecheney:
@asbjxrn i'm getting on a plane shortly and will be out of contact for a few days. In my absence @davidkaya has my proxy to resolve this issue.

davecheney:
Thanks

PR #259, opened 2015-07-06
Allow git supported repos for code.google.com

sheenobu:


davecheney:
Thanks for your contribution. I don't know how much longer google code will be around, but this support cannot hurt.

PR #261, opened 2015-07-07
Fix typo in vendor fetch doc

ulfnielsen:
This just kept attracting my eye sorry. "supplie" -> "supplied"


davecheney:
Please gofmt this change then I will submit it. Thanks.

ulfnielsen:
Done, but kinda strange having to run gofmt on a file coming from go generate and which includes a "do not edit" comment.

davecheney:
Yes, it sucks a bit. If you'd like to track down the two bits of whitespace
in the template, then we won't have to do this extra step.

ulfnielsen:
I've removed the whitespace as well (missed it the first time around) and removed the leading newline in the template strings.

davecheney:
LGTM. Thanks

PR #265, opened 2015-07-11
slightly improved output when running gb-vendor directly

gpaul:
My instinct after go get'ing stuff is to run '-help' on all the binaries that appeared that I didn't expect. In the case of gb-vendor, that resulted in some confusing output about 'project root'. 

I would have found the usage text much more useful as it starts with 'gb vendor', immediately suggesting that gb-vendor is run by the gb command and shouldn't be run by hand.

This patch is purely cosmetic.

### DIFF
--- a/cmd/gb-vendor/main.go
+++ b/cmd/gb-vendor/main.go
@@ -31,13 +31,6 @@ var commands = []*cmd.Command{
 }
 
 func main() {
-	root, err := cmd.FindProjectroot(projectroot)
-	if err != nil {
-		gb.Fatalf("could not locate project root: %v", err)
-	}
-	project := gb.NewProject(root)
-	gb.Debugf("project root %q", project.Projectdir())
-
 	args := os.Args[1:]
 	if len(args) < 1 || args[0] == "-h" {
 		fs.Usage()
@@ -48,6 +41,13 @@ func main() {
 		help(args[1:])
 		return
 	}
+
+	root, err := cmd.FindProjectroot(projectroot)
+	if err != nil {
+		gb.Fatalf("could not locate project root: %v", err)
+	}
+	project := gb.NewProject(root)
+	gb.Debugf("project root %q", project.Projectdir())
 
 	for _, command := range commands {
 		if command.Name == args[0] && command.Runnable() {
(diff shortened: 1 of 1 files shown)


cmd/gb-vendor/main.go
```diff
@@ -31,23 +31,27 @@ var commands = []*cmd.Command{
...
+
+	switch {
+	case len(args) < 1, args[0] == "-h", args[0] == "-help":
 		fs.Usage()
 		os.Exit(1)
-	}
-
-	if args[0] == "help" {
+	case projectroot == "":
+		gb.Warnf("don't run this binary directly, it is meant to be run as 'gb vendor ...'")
```
davecheney: Please make this a fatal and drop the next line, the goal is not to encourage the user to continue to use it incorrectly.

davecheney:
LGTM with minor nits. Please fix them up and i'll merge this.

cmd/gb-vendor/main.go
```diff
@@ -31,23 +31,26 @@ var commands = []*cmd.Command{
...
+	switch {
+	case len(args) < 1, args[0] == "-h", args[0] == "-help":
 		fs.Usage()
 		os.Exit(1)
-	}
-
-	if args[0] == "help" {
+	case projectroot == "":
+		gb.Fatalf("don't run this binary directly, it is meant to be run as 'gb vendor ...'")
+	case args[0] == "help":
```
davecheney: This logic never fires,

zapf(~/src/github.com/constabulary/gb/cmd) % gb-vendor help
FATAL don't run this binary directly, it is meant to be run as 'gb vendor ...'

I think if you move it above the `case projectroot == ""` block it will work

gpaul: I think the idea is that if you run gb-vendor directly it should tell you to run it as 'gb vendor ...' It seems like that should hold for 'gb-vendor help', too. The case in question fires when you run it properly, as 'gb vendor help'.

That said, it's always nicer to be greeted by help text instead of a fatal error log. I've swapped the cases as suggested - it works.

davecheney:
LGTM. Thanks for fixing this.

gpaul:
:)

PR #269, opened 2015-07-13
cmd/gb: add gb info subcommand

### LINKED ISSUE #267: cmd/gb: new subcommand, info
This is a request from @freeformz who needs a way to tell if a repo root is a gb project.

My solution is to create a new subcommand `info`, which has output similar to `git status` and an exit code that will be non zero if the repo does not small like a gb project.

davecheney:
Fixes #267

This PR adds a new subcommand, `info`. This returns some (hopefully) pertinent
information about the project, and an error code to indicate if the path is part
of a gb project or not.

	zapf(~/devel/demo4) % gb info
	GB_PROJECT_DIR="/Users/dfc/devel/demo4"
	GB_SRC_PATH="/Users/dfc/devel/demo4/src:/Users/dfc/devel/demo4/vendor/src"
	zapf(~/devel/demo4) % echo $?
	0

	zapf(~/devel/demo9) % gb info
	FATAL could not locate project root: could not find project root in "/Users/dfc/devel/demo9" or its parents
	zapf(~/devel/demo9) % echo $?                                                                         
    1

### DIFF
--- a/cmd/gb/alldocs.go
+++ b/cmd/gb/alldocs.go
@@ -14,6 +14,7 @@ The commands are:
         doc         show documentation for a package or symbol
         env         print project environment variables
         generate    generate Go files by processing source
+        info        info returns information about this project
         list        list the packages named by the importpaths
         plugin      run a plugin
         test        test packages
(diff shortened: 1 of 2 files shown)


davecheney:
@freeformz PTAL. The stdout of this command is not well defined -- perhaps it should be json, but I figured that would be secondary to your use case.

tianon:
FWIW, +1 on having a shell sourceable version of this output still if it ever switches to JSON officially. :smile:

davecheney:
@tianon same. I think if folks need machine consumable json then , `gb info -json` will be the order of the day.

tianon:
:cake: :heart:

PR #278, opened 2015-07-15
Fix vendor introduced with #275

tarfu:
gb vendor needs the new syntax for NewProject to work.


davecheney:
Thanks. I will add some tests so I don't break this again.

PR #303, opened 2015-08-14
cgo pkg-config fixes

sheenobu:
This fixes two issues with cgo support:

 * In go 1.4, the pkg-config values must be passed to cgo after the -- (might be the same for go 1.5). CGO_CFLAGS seems to be ignored.
 * Both versions require the cflags/pkg-config to be passed to rungcc1.

Repo which replicates this issue with working output vs current gb master output: https://github.com/sheenobu/gb-cgo-bug

### DIFF
--- a/cgo14.go
+++ b/cgo14.go
@@ -24,7 +24,7 @@ func cgo(pkg *Package) ([]ObjTarget, []string) {
 	cgoCFLAGS = append(cgoCFLAGS, pcCFLAGS...)
 	cgoLDFLAGS = append(cgoLDFLAGS, pcLDFLAGS...)
 
-	if err := runcgo1(pkg, cgoCFLAGS, cgoLDFLAGS); err != nil {
+	if err := runcgo1(pkg, cgoCFLAGS, pcCFLAGS, cgoLDFLAGS); err != nil {
 		return fn(ErrTarget{err})
 	}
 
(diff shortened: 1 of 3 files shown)


cgo14.go
```diff
@@ -24,7 +24,7 @@ func cgo(pkg *Package) ([]ObjTarget, []string) {
 	cgoCFLAGS = append(cgoCFLAGS, pcCFLAGS...)
 	cgoLDFLAGS = append(cgoLDFLAGS, pcLDFLAGS...)
 
-	if err := runcgo1(pkg, cgoCFLAGS, cgoLDFLAGS); err != nil {
+	if err := runcgo1(pkg, cgoCFLAGS, pcCFLAGS, cgoLDFLAGS); err != nil {
```
davecheney: Why is this necessary, pcFLAGS are already included in cgoCFLAGS

davecheney:
Thanks for your fix and the reproduction, i'll include this integration tests.

cgo14.go
sheenobu: cgo is not respecting the contents of cgoCFLAGS. Additionally, cgoCFLAGS is CFLAGS + "-g -02" + pcCFLAGS. -g -02 are not valid arguments for cgo which means we need to include pcCFLAGS separately when building the full cgo command.

sheenobu:
@davecheney Great, thank you.

davecheney:
@sheenobu please take a look at the integration test failures.

davecheney:
@sheenobu i'm going to take a look at this now. I've created an integration test for sdl2 and can see the problem that we're not getting the right flags out of pkg-config into runcgo1

sheenobu:
Looks like only half of this PR was correct. Replaced by #304

Issue #319, opened 2015-08-23
gb build failed for go1.5


linuxerwang:
I upgraded to go1.5, rebuilt the whole gb pkg/cmds, and "gb build" gave error:

FATAL command "build" failed: fork/exec /usr/share/go-golang/pkg/tool/linux_amd64/6g: no such file or directory

Also tried to rebuild gb with build tag (I know it's redundant):

go install --tags 'go1.5' github.com/constabulary/gb/cmd/gb
go install --tags 'go1.5' github.com/constabulary/gb/cmd/gb-vendor

But still "gb build" failed with the above error.

Thanks.

davecheney:
It's likely that gb didn't get rebuilt, you can try adding the `-v` flag to go install, and seeing if it's actually doing anything.

Personally, I recommend removing `$GOPATH/pkg` and running `go install -v .../cmd/gb`

linuxerwang:
Same issue. I even manually deleted .a and binary.

rm -rf pkg/linux_amd64/github.com/constabulary/gb/*
rm -rf bin/gb*
go install -v .../cmd/gb
go install -v .../cmd/gb-vendor

davecheney:
Are you saying that go install did nothing ? Can you please include the
full output of what you see ?

davecheney:
`/usr/share/go-golang/pkg/tool/linux_amd64/6g` is not a common installation path for Go. Where did you get this go install from ? Do you or have you set GOROOT ? If so, please do not, and repeat the procedure.

davecheney:
Can you please include the ouput of `which go`, is it the version you expect ?

linuxerwang:
gb and gb-vendor were installed correctly, I can run "gb vendor fetch" without problem.  The problem is when running "gb build".

I build my go1.5 from source code. And there is no 6g under /usr/share/go-golang/pkg/tool/linux_amd64/6g. I think go1.5 removed all 6g/6l etc. according to the release notes: https://golang.org/doc/go1.5#compiler_and_tools

No I did not set GOROOT.

davecheney:
Thank you for confirming. My suspicion is that your go installation is not correct, and that incorrect value is being encoded into gb when it is compiled (because it must know this value to know where to find the compiler).

The fact that gb is using the go 1.4 behavior (looking for `6g`, not `compile`) also lead me to believe that the go installation on your computer is not correct. 

How have you installed go ? If you uninstall go and then issue `go`, does it suceed ? What are the values on your $PATH ?

[...]

PR #321, opened 2015-08-23
cmd/gb: verify the Go install has not changed

davecheney:
Update #319 

This PR adds a check to attempt to diagnose gb failures due to a mismatch between the version of Go that compiled gb, and the version of Go that exists on the machine when gb is invoked.

I expect it will cause some false positives, and if so, we'll address them or remove the check.
```
% gb build
FATAL Go version does not match: gb was compiled with "devel +fa87cf8 Wed Aug 19 04:36:55 2015 +0000", installed Go version from /home/dfc/go is "devel +8ca7856 Sat Aug 22 18:39:29 2015 +0000"
```

### DIFF
--- a/cmd/gb/main.go
+++ b/cmd/gb/main.go
@@ -2,8 +2,11 @@ package main
 
 import (
 	"flag"
+	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
+	"strings"
 
 	"github.com/constabulary/gb"
 	"github.com/constabulary/gb/cmd"
@@ -46,6 +49,9 @@ func main() {
 		help(args[2:])
 		return
 	}
+
+	verifyGoVersion()
+
 	command, ok := commands[name]
 	if (command != nil && !command.Runnable()) || !ok {
 		if _, err := lookupPlugin(name); err != nil {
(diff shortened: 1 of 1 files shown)


tianon:
Not sure if it warrants opening a new issue (although happy to do so if you think I ought), but this causes failures on Gentoo due to `VERSION` not being installed in `GOROOT`:
```console
$ go version
go version go1.4.2 linux/amd64
$ go env GOROOT
/usr/lib/go
$ ls -lAFh /usr/lib/go
total 24K
drwxr-xr-x  3 root root 4.0K Apr  4 13:24 4.8.4/
drwxr-xr-x  8 root root 4.0K Mar 19 23:02 doc/
drwxr-xr-x  3 root root 4.0K Mar 19 23:02 include/
drwxr-xr-x  3 root root 4.0K Dec 16  2013 lib/
drwxr-xr-x  5 root root 4.0K Oct  9  2013 pkg/
drwxr-xr-x 46 root root 4.0K Mar 19 23:02 src/
```

davecheney:
I knew there would be one distro that would mes this up.

I think it is a mistake that gentoo filter the installed contents of
GOROOT. What are the chances of getting you to raise the bug against them,
not gb?

tianon:
Heh, I'm not opposed, but to be honest also not _completely_ convinced.  Is there a good reference anywhere about what's expected to live in `GOROOT`?  I don't think I've ever seen one -- just assumptions here and there (like the `GOROOT_BOOTSTRAP` code assuming that `GOROOT/bin/go` exists, for example, which Debian had to work around for `gccgo`).  It is worth noting that Debian does install the `VERSION` file in `GOROOT`, but I've always understood `go version` to be the canonical method of fetching the "current Go version" (or `runtime.Version`).

Do you think there'd be any chance at all of convincing Go upstream to clearly document the things expected to exist in `GOROOT` so we could use that as leverage?  As one of the packagers for Go in Debian, I know I'd personally really appreciate an official document like that, and I don't think I'd be alone in appreciation of that (and that would make it really hard to argue that this is a `gb` bug and not a Go packaging bug).

davecheney:
My hope is I can get upstream to adopt this method of detecting go install
mismatches as people setting GOROOT and forgetting it has been a continual
problem for the last four years.

If I am successful in that, the proposal will include the requirement that
VERSION is present.

tianon:
That would definitely be very interesting! :+1:

I've done a little digging, and found this gem: https://bugs.debian.org/693186

Apparently, `go tool dist` doesn't work without `$GOROOT/VERSION`. :smiling_imp:

tianon:
https://bugs.gentoo.org/show_bug.cgi?id=558620 :wink: :heart:

davecheney:
Thanks. I appreciate that someone else gets to hold this hot potato for a
time.

PR #339, opened 2015-09-06
Remove hard-coded -m64 references

### LINKED ISSUE #338: -m64?
What's the rationale for always including `-m64` in `cgo.go`?  I searched the issues and couldn't find a relevant discussion, but if there's one I missed, please feel free to redirect me! :+1:

(Asking because of https://bugs.debian.org/798135, which is the report that we fail to build on 32 bit arches like armel, armhf, and i386.)

If it's superfluous, I'd be happy to send a PR to remove it, but I'm guessing there's something more subtle I missed, hence opening an issue first. :smile:

tianon:
Fixes #338 :+1:


davecheney:
LGTM.

PR #355, opened 2015-09-09
Fix transposed context values.

ChrisHines:
Fixes bug introduced during recent context refactor in commit https://github.com/constabulary/gb/commit/4097af980ccf60c79b4c8350369192982f744bad.


davecheney:
Thanks. I'll submit this once wercker goes green.

PR #359, opened 2015-09-10
Improve the hint when gb failed to fetch the vendor package.

hyper-carrot:
As it said in title.

### DIFF
--- a/vendor/imports.go
+++ b/vendor/imports.go
@@ -83,11 +83,10 @@ func fetchMetadata(scheme, path string, insecure bool) (io.ReadCloser, error) {
 				return resp.Body, nil
 			}
 		}
+	default:
+		return nil, fmt.Errorf("unknown remote protocol scheme: %q", scheme)
 	}
-	if err != nil {
-		return nil, fmt.Errorf("fail to access url %q", url)
-	}
-	return nil, fmt.Errorf("unknown remote protocol scheme: %q", scheme)
+	return nil, fmt.Errorf("fail to access url %q", url)
 }
 
 // ParseMetadata fetchs and decodes remote metadata for path.
--- a/vendor/imports.go
+++ b/vendor/imports.go
@@ -51,19 +51,26 @@ func ParseImports(root string) (map[string]bool, error) {
 // FetchMetadata fetchs the remote metadata for path.
 func FetchMetadata(path string, insecure bool) (io.ReadCloser, error) {
 	schemes := []string{"https", "http"}
+	var err error
+	var r io.ReadCloser
 	for _, s := range schemes {
-		if r, err := fetchMetadata(s, path, insecure); err == nil {
+		if r, err = fetchMetadata(s, path, insecure); err == nil {
 			return r, nil
 		}
+	}
+	if err != nil {
+		return nil, err
 	}
 	return nil, fmt.Errorf("unable to determine remote metadata protocol")
 }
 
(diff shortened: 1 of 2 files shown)


davecheney:
LGTM. Thank you.

This is one area that does have some tests, could you please try adding a test (you may need to extend the test set up) here https://github.com/constabulary/gb/blob/master/vendor/imports_test.go#L27

hyper-carrot:
@davecheney ok, I will try.

hyper-carrot:
@davecheney  done, but waiting for Wercker.

hyper-carrot:
@davecheney I had updated it, I think that these hints more better than before.

davecheney:
Thanks. I'll check it this evening and merge it

hyper-carrot:
It's my pleasure. :-)

vendor/imports.go
```diff
@@ -50,20 +50,33 @@ func ParseImports(root string) (map[string]bool, error) {
 
 // FetchMetadata fetchs the remote metadata for path.
 func FetchMetadata(path string, insecure bool) (io.ReadCloser, error) {
-	schemes := []string{"https", "http"}
+	var schemes []string
```
davecheney: please try this instead

var schemes = []string { "https" }
if insecure {
     schemes = append(schemes, "http")
}

davecheney:
The last change set is not lgtm. Please revert it and we can take it up in a future PR.

hyper-carrot:
OK. I will do.

But, in fact, It always make gb print out `fail to access url "http://any.inaccessible.server/the.project?go-get=1"` (without `https`) when `insecure == false` with `schemes := []string{"https", "http"}`. I think that it is not correct.

davecheney:
@hyper-carrot sure, i expect that will cause one of the tests you added to fail, that was probably why the error was being discarded in the first place. 

ok, then please move the insecure check out of fetchMetadata. That is, only call fetchMetadata with a permitted scheme, don't handle it in FetchMetadata and fetchMetadata

hyper-carrot:
@davecheney done. The TestFetchMetadata() will fail.  What to do next?

davecheney:
@hyper-carrot sadly you have to make the tests pass. This is a good change, but the logic before ignored the error hitting a possibly invalid url for this reason.

hyper-carrot:
@davecheney  OK, sorry, I reset it.

hyper-carrot:
@davecheney Done. And maybe FetchMetadata() can be modified like this:

```go
func FetchMetadata(path string, insecure bool) (io.ReadCloser, error) {
	schemes := []string{"https"}
	if !insecure {
		gb.Infof("skipping insecure protocol for %q", path)
	} else {
		schemes = append(schemes, "http")
	}
	var err error
	var r io.ReadCloser
	for _, s := range schemes {
		if r, err = fetchMetadata(s, path); err == nil {
			return r, nil
		}
	}
	if err != nil {
		return nil, err
	}
	return nil, fmt.Errorf("unable to determine remote metadata protocol")
}

func fetchMetadata(scheme, path string) (io.ReadCloser, error) {
	url := fmt.Sprintf("%s://%s?go-get=1", scheme, path)
	var err error
	var resp *http.Response
	switch scheme {
	case "https", "http":
		resp, err = http.Get(url)
		if err == nil {
			return resp.Body, nil
		}
	default:
		return nil, fmt.Errorf("unknown remote protocol scheme: %q", scheme)
	}
	return nil, fmt.Errorf("fail to access url %q", url)
}
```

Issue #14, opened 2015-04-29
should the gb repo be a project?


ccmtaylor:
Reading through `README.md` and `getting-started.md`, I got a bit confused whether projects are meant to be checked into version control, and how the different directory structure would interact with making a project `go get`able.

So I thought that since gb is proposing the new structure, and it seems to fit the definition of a project (i.e. it has a package main, and isn't designed to be consumed by other code), it would make sense for the gb repo to be structured as a project itself.

I'm not sure how to go about this though. If I understand correctly, this would only entail moving the code to `gb/src`. This creates a chicken-and-egg problem, though, since `cmd/path.go` and `cmd/gb/main.go` both import `github.com/constabulary/gb`. When I build with `go build`, this would now need to be `import "github.com/constabulary/gb/src"`, and building with `gb` would probably expect the code to live under `src/github.com/constabulary/gb`, right?

davecheney:
In an ideal world; yes. At the moment, there is a bootstrapping problem, so unless `gb` grows a stack of dependencies, I don't think this is super urgent.

davecheney:
You can try if you like, this is a low priority item for me at the moment.

skelterjohn:
I think it's appropriate to leave it go gettable permanently. Go get is a great way to fetch code. Dave just has to keep the transitive deps low or zero.

andrewchambers:
This will be fine once there is a hosted prebuilt binary to download.

ccmtaylor:
Closing due to age

PR #361, opened 2015-09-11
add version command

willf:
Wanted to be able to ask the `gb` command what its current version was, but I see no way to do so. This adds a `version` command to do this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 !*
 testdata/bin/
+*.bk
--- a/cmd/gb/version.go
+++ b/cmd/gb/version.go
@@ -13,7 +13,8 @@ func init() {
 		UsageLine: `version`,
 		Short:     "show current version of gb",
 		Run: func(ctx *gb.Context, args []string) (err error) {
-			fmt.Println("0.1.1")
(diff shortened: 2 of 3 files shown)


davecheney:
I;m fine with adding a version command, but I'm sure i'll forget to update that constant. Is there any way to automate this ?

willf:
I added a 'version.sh' script which will modify the version.go file. It's possible to add this as a git hook, but I think you have to do it on your side.

freeformz:
We faced the same problem with tools/godep and just agreed to bump the version with any code changes to master. Yes that sucks BTW.

davecheney:
Thank you for this proposal. Sadly I am not going to merge this at this time.

I do want a version command, but I'm not interested in adding one until it is completely automatic. 

By this I mean, the value reported by gb version should match the tag or revision that gb was built from. I don't think we can do that today with the go tool, and simulating it with a constant would mean releasing from a branch, or a more involved release process involving setting the version to the release version comitting, tagging, then changing it back to some opaque devel version.

At the rate gb is iterating I don't think this overhead is worthwhile. Possibly in the future it will be, maybe when we get to self hosting #14.

PR #364, opened 2015-09-15
Improve hint & steps in the metadata fetch process.

hyper-carrot:
sorry, it seems a bit late.

PR #365, opened 2015-09-15
vendor: improve hint & steps in the metadata fetch process

davecheney:
From #364

### DIFF
--- a/vendor/imports.go
+++ b/vendor/imports.go
@@ -52,7 +52,7 @@ func ParseImports(root string) (map[string]bool, error) {
 func FetchMetadata(path string, insecure bool) (io.ReadCloser, error) {
 	schemes := []string{"https"}
 	if !insecure {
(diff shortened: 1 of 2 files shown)


vendor/imports.go
```diff
@@ -50,11 +50,16 @@ func ParseImports(root string) (map[string]bool, error) {
 
 // FetchMetadata fetchs the remote metadata for path.
 func FetchMetadata(path string, insecure bool) (io.ReadCloser, error) {
-	schemes := []string{"https", "http"}
+	schemes := []string{"https"}
```
davecheney: @hyper-carrot  I'm sorry this isn't right. Now the user will always get a message saying "skipping insecure propocol", even if it is never required that the server hit the insecure protocol.

I'll try to fix this

hyper-carrot:
@davecheney OK, I see.

davecheney:
Something horrible happened to this PR. I'll try again from scratch

PR #370, opened 2015-09-16
add service.v1 integration test

davecheney:


davecheney:
Sigh - another thing that wercker doesn't support

tianon:
If you'd be interested in trying Travis, I'd be willing to play with setting up a `.travis.yml` build matrix for you. :+1:

davecheney:
Sure. Go for it.

tianon:
:metal: will play! :heart:

davecheney:
I used wercker because it gives pre commit testing. I don't know much about
travis, but pre commit testing is key.

mmlb:
You can have precommit testing + travis if you throw in another tool/service in the mix: https://github.com/barosl/homu

davecheney:
This is sounding less and less attractive ...

mmlb:
:D

tianon:
Heh, yeah, it doesn't fit your needs.  Perhaps I'll just play with werker
instead and see if I can help debug. :+1:

davecheney:
I think the solution to the wercker problem is to start with a different
"box", one that is based on a more complete os image.

It's not a big deal, all those integration tests are in the repo any way,
even if they cannot all be run through CI directly.

PR #371, opened 2015-09-16
Fix: Clean working directory

mopemope:
`gb test` failed leak working directory.
`log.Fatal` call `os.Eixt`. deferred functions are not run.


davecheney:
Thank you. This will fix a lot of leaks, test's fail often.

PR #380, opened 2015-09-17
experiment: can we specify multiple boxes and build stages

davecheney:


davecheney:
Nope, turns out we cannot, not without switching to the docker based version.

tianon:
I've been playing with wercker and making progress, but I didn't realize you weren't using the Docker workers so my work so far might be kind of moot. :innocent:

davecheney:
We can use the docker version if you like. Here are the list of nice to
haves

- multiple versions of Go, 1.4, 1.5 and tip if possible (stretch goal)
- smells more like ubuntu 14.04, so we have all the -dev packages that the
integration tests need

tianon:
Ok, I'll keep playing then! :+1:

So far I'm working on just getting all the tests running (including that new one you weren't able to add in #370 and the one that is in the file but is commented out), but it's good to have a list of the things you'd really like to see so I can focus specifically on those! :smile:

davecheney:
I'll settle for a fixed version of Go (1.4 is fine), and more integration
tests over running multiple versions of Go

PR #383, opened 2015-09-17
vendor: give better error when fetch fails on insecure report import

### LINKED ISSUE #358: Can not fetch the package with a insecure protocol from my custom import path
Please see contents below:
```bash
$ gb vendor fetch go.wepiao.com/base_lib/log
skipping insecure protocol: http://go.wepiao.com/base_lib/log?go-get=1
FATAL command "fetch" failed: unable to determine remote metadata protocol
```

As you can see, It is failed to get the package from our git server, but `go get -insecure go.wepiao.com/base_lib/log` is OK.

I think that gb should support `-insecure ` flag.

davecheney:
Fixes #358

### DIFF
--- a/vendor/imports.go
+++ b/vendor/imports.go
@@ -50,43 +50,36 @@ func ParseImports(root string) (map[string]bool, error) {
 
 // FetchMetadata fetchs the remote metadata for path.
 func FetchMetadata(path string, insecure bool) (io.ReadCloser, error) {
-	schemes := []string{"https", "http"}
-	var err error
-	var r io.ReadCloser
-	for _, s := range schemes {
-		if r, err = fetchMetadata(s, path, insecure); err == nil {
-			return r, nil
-		}
+	// try https first
+	r, err := fetchMetadata("https", path)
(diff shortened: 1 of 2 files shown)


hyper-carrot:
got it!

hyper-carrot:
Er...I don't think the error message is uniform. 

What about this?

```
func FetchMetadata(path string, insecure bool) (rc io.ReadCloser, err error) {
        defer func() {
		if err != nil  {
			err = fmt.Errorf("unable to determine remote metadata protocol: %s", err)
		}
        }()
	// try https first
	rc, err = fetchMetadata("https", path)
	if err == nil {
		return
	}
	// try http if supported
	if insecure {
		rc, err = fetchMetadata("http", path)
		return
	} else {
		log.Infof("skipping insecure protocol: %v", "http")
	}
	return
}
```

davecheney:
@hyper-carrot, please feel free to send a PR

hyper-carrot:
@davecheney okok ;)

PR #388, opened 2015-09-17
Switch the README badge to show stats for master instead of all builds

tianon:


davecheney:
LGTM. So that's how you do it!

tianon:
Heh, to be honest I went to "share" in their interface which has a nice "Master" radio button now that made it obvious for me -- probably wasn't there when you first added this. :smile:

PR #389, opened 2015-09-17
cmd/gb: add build tags

### LINKED ISSUE #81: -tags support
Does gb plan to support custom build constraints? ie `go build -tags dev` mentioned here: http://golang.org/pkg/go/build/#Context

davecheney:
Add support for build tags. Build tags modify the package resolution logic, a la build constraints, build tags also affect staleness computation.

Packages built with tags in effect will be stored in their own package cache. Binaries built with with tags in effect will have the tags appended as a suffix.

Fixes #81 

- [x] extend stringVars to other flags in cmd/gb
- [x] add tests

### DIFF
--- a/build_test.go
+++ b/build_test.go
@@ -17,8 +17,11 @@ import (
 func TestBuild(t *testing.T) {
 	log.Verbose = false
 	defer func() { log.Verbose = false }()
+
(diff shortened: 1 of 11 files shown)


tbruyelle:
Nice, I was waiting for this

paxan:
:+1:

PR #390, opened 2015-09-17
Add ".exe" to the TestPackageBinfile test for windows binaries

tianon:
Fixes:

```
--- FAIL: TestPackageBinfile (0.00s)
	package_test.go:56: (b).Binfile(): want C:\gopath\src\github.com\constabulary\gb\testdata\bin\b, got C:\gopath\src\github.com\constabulary\gb\testdata\bin\b.exe
```

Not sure if this is the right fix, but I'll let you judge me on that. :smile:

### DIFF
--- a/package_test.go
+++ b/package_test.go
@@ -52,6 +52,9 @@ func TestPackageBinfile(t *testing.T) {
 		}
 		got := pkg.Binfile()
 		want := filepath.Join(ctx.Bindir(), tt.want)
+		if pkg.gotargetos == "windows" {
+			want += ".exe"
+		}
 		if want != got {
 			t.Errorf("(%s).Binfile(): want %s, got %s", tt.pkg, want, got)
 		}


package_test.go
```diff
@@ -52,6 +52,9 @@ func TestPackageBinfile(t *testing.T) {
 		}
 		got := pkg.Binfile()
 		want := filepath.Join(ctx.Bindir(), tt.want)
+		if pkg.gotargetos == "windows" {
```
tianon: I wasn't entirely sure what `tt.goos` is for, but it's not set to `"windows"` so I went for `pkg.gotargetos` which was appropriately set to `"windows"`. :confused:

tianon: (Especially since `tt.goos` and `tt.goarch` aren't actually used in this block anywhere I can see. :smile:)

davecheney: they were supposed to feed into NewContext, via gb.GOOS, and gb.GOARCH, but they aren't hooked up yet. No matter.

davecheney:
Yup, LGTM.

tianon:
:tada:

PR #395, opened 2015-09-18
vendor: give better error when fetch fails on insecure report import (improvement 1)

hyper-carrot:
It will print `unable to determine remote metadata protocol: failed to access url "..."` when fetch fails on insecure report import, no matter what insecure is set or not.

### DIFF
--- a/vendor/imports.go
+++ b/vendor/imports.go
@@ -49,23 +49,25 @@ func ParseImports(root string) (map[string]bool, error) {
 }
 
 // FetchMetadata fetchs the remote metadata for path.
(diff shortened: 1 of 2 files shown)


davecheney:
@hyper-carrot somehow this PR is broken, possibly your editor or environment has mangled the whitespace in the werker yaml file, although github cannot detect it.

hyper-carrot:
@davecheney  But I don't have edited werker.yml, and I haven't found any difference.  I will try to fix.

tianon:
Might just need a rebase against latest master. :)

- Tianon

davecheney:
It's ok, i've replaced this PR with another.

PR #397, opened 2015-09-18
Add initial appveyor.yml

tianon:
See https://ci.appveyor.com/project/tianon/gb for where I've got this setup on my fork.  I've managed to get the build down to ~2m. :smile:

If you like what you see, setting up AppVeyor should be as simple as creating an account, authorizing them for GitHub, and adding "constabulary/gb" as a project.  I'd also recommend turning on the "ignore branches without `appveyor.yml`" setting, but none of the other settings should be necessary (since they're mostly specified here and having an `appveyor.yml` overrides the entire UI settings anyhow).

I _think_ we can get the integration tests running here too, but that'll take a little more refactoring over in the other repo (especially the `setup.bash` files).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,41 @@
+version: build-{build}.{branch}
+
+clone_folder: C:\gopath\src\github.com\constabulary\gb
+shallow_clone: true # for startup speed
+
+environment:
+  GOPATH: C:\gopath
+
+platform:
+  - x64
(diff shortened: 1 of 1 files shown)


tianon:
I'd note additionally that you'll want to make sure you choose the "free" plan -- it's totally gratis for open source. :smile:

davecheney:
Nice one @tianon !

PR #402, opened 2015-09-19
Doc tidying: add missing flag docs for list command, line wrapping

ches:
I noticed that the usage line for `list` showed a few flags that weren't explained. While adding those explanations, I also made line wrapping for some help output that had bugged me on narrower terminals (`build`) consistent with other commands that were nice and pretty, and generally tried to make all the source formatting consistent.

I'm new to gb and just started browsing the source, so there may be some discrepancies I've missed fixing (not sure yet if all the `build` flags are correctly documented), but I think this is a good start.

### DIFF
--- a/cmd/gb/alldocs.go
+++ b/cmd/gb/alldocs.go
@@ -16,13 +16,13 @@ The commands are:
         generate    generate Go files by processing source
         info        info returns information about this project
         list        list the packages named by the importpaths
-        plugin      run a plugin
         test        test packages
 
 Use "gb help [command]" for more information about a command.
 
 Additional help topics:
 
+        plugin      plugin information
         project     gb project layout
 
 Use "gb help [topic]" for more information about that topic.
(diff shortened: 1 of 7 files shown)


davecheney:
Thanks for the clean up. 

One thing to remember, after changing the docstrings you need to rebuild cmd/gb, then in cmd/gb run go generate to regenerate alldocs.go (see the top of the file). I've done that for you when merging this change.

Thanks

Dave

ches:
Thanks, I did notice that a `generate` needed to be run but I was sort of guessing at how exactly—I just ran `go generate cmd/gb/alldocs.go`.

I'd like to remove the guesswork by automating this or at least putting a note in the contributing section of the README. So should the steps be as follows?

```sh
$ cd cmd/gb
$ go build
$ go generate
```

When I do this on master currently I'm getting a diff that reverses a99a196d358230, so it feels like I must be missing something…

davecheney:
@ches sgtm, please open an issue or pull request

PR #411, opened 2015-09-23
Move ".exe" appending logic into ctx.Suffix()

tianon:
This also updates the appropriate places in the code to use `ctx.Suffix()` where they weren't previously (and actually, nothing but `gb info` was using it).

One notable place that _doesn't_ change here is `gc.go`, since it's using `gohostos` to find existing tool locations, not dealing with output locations (`gotargetos`).

### DIFF
--- a/cmd/test_test.go
+++ b/cmd/test_test.go
@@ -143,10 +143,10 @@ func TestTestPackages(t *testing.T) {
 		err     error
 	}{{
 		pkgs:    []string{"a", "b", "c"},
-		actions: []string{"run: [$WORKDIR/a/testmain/_test/a$EXE.test]", "run: [$WORKDIR/b/testmain/_test/b$EXE.test]", "run: [$WORKDIR/c/testmain/_test/c$EXE.test]"},
+		actions: []string{"run: [$WORKDIR/a/testmain/_test/a$SUFFIX.test]", "run: [$WORKDIR/b/testmain/_test/b$SUFFIX.test]", "run: [$WORKDIR/c/testmain/_test/c$SUFFIX.test]"},
 	}, {
 		pkgs:    []string{"cgotest", "cgomain", "notestfiles", "cgoonlynotest", "testonly", "extestonly"},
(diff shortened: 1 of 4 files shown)


tianon:
It was really bothering me that we had `.exe` appending logic in so many places. :innocent:

davecheney:
This feeds wrong sorry, Suffux is composed of goos, goarch and tags, it shouldn't have anything to do with the nasty hack required for windows. The rest of the cleanup is good thou.

tianon:
I'm not sure I follow -- right now in `master`, `ctx.Suffix()` is essentially `if ctxString() is not empty, return "-" + ctxString()`, but when we build we do `if cross-compiling, append "-" + ctxString() to our binary name`, which will be the same result?  So you're saying using tags + cross-compile will actually be kind of funky as-is with `master`?

davecheney:
it really bothers me that windows requires this gross haxk

tianon:
(and besides that, the _only_ place `ctx.Suffix()` is currently used is in `gb info` for an output field called `GB_BIN_SUFFIX`, so I was confused as to what that means)

tianon:
Yeah, nasty windows. :cry:

tianon:
Would you prefer a new function for this, or should I modify the way `ctx.Suffix()` works to not include tags info?

tianon:
(and maybe rename it to `BinSuffix` or something so it's clear?)

davecheney:
Neither. I don't want ctxString or Suffix to have any .exe entry, even on
windows. That should be handed in binffile/Binfile

davecheney:
There should be only one place where the names of binaries are computed,
even for test binaries

tianon:
Ok, that's fair, but does mean our tests will need to keep some of the same logic (like they have today).

Do you think it'd make sense to have a function somewhere to encapsulate our `if` statement for going from `OS` to "binary extension"? (ie, something like `func binExtension(os string) string { if os == "windows" { return ".exe" } return "" }` so we could just `+= binExtension(ctx.gotargetos)` or `+= binExtension(ctx.gohostos)`)

davecheney:
I don't think this is worth a big effort. There are only two places that generate build.go via link and cmd/test.go, both of those should be using some kind of helped to generate the binary name. If not, we add a helper and a unit test. I don't think this needs to be more complex than that.

PR #415, opened 2015-09-25
Fix Binfile for all permutations of GOOS/GOARCH/Tags

### LINKED ISSUE #414: gb build: cross compile + tags leads to duplication in filename
If I do something like `GOOS=windows GOARCH=386 gb build -tags lol`, my final binary name is `cmd-windows-386-lol-lol.exe` instead of just `cmd-windows-386-lol.exe` (due to `ctxString` being added during cross-compile and `buildtags` being included in that).

My thoughts for fixing this would be to use `Suffix()` for both instead, which prepends the `-` appropriately, but the problem there is that it always adds both `gotargetos` and `gotargetarch` regardless of cross-compile.  Since `Suffix()` is _only_ used for `GB_BIN_SUFFIX` in `gb info` (which on my machine is always `GB_BIN_SUFFIX="-linux-amd64"` unless cross-compiling or using tags), I'm confused about this discrepancy, since the name suggests that it's the suffix that'll be appended to the names of binaries output by `gb build`, and thus should be the empty string for no tags and no cross-compile, right?  If this sounds like I'm on the right track, I'm happy to resurrect those parts of #411 as appropriate, but if I'm on the wrong track then I'd really appreciate some insight into what `GB_BIN_SUFFIX` is useful for. :innocent:

Essentially:
```diff
diff --git a/context.go b/context.go
index feac334..0d4a766 100644
--- a/context.go
+++ b/context.go
@@ -167,7 +167,12 @@ func (c *Context) Pkgdir() string {
 // Suffix returns the suffix (if any) for binaries produced
 // by this context.
 func (c *Context) Suffix() string {
-	suffix := c.ctxString()
+	v := []string{}
+	if c.isCrossCompile() {
+		v = append(v, c.gotargetos, c.gotargetarch)
+	}
+	v = append(v, c.buildtags...)
+	suffix := strings.Join(v, "-")
 	if suffix != "" {
 		suffix = "-" + suffix
 	}
diff --git a/package.go b/package.go
index ed26ccb..b30a534 100644
--- a/package.go
+++ b/package.go
@@ -91,15 +91,7 @@ func (pkg *Package) Binfile() string {
 		target = filepath.Join(pkg.Bindir(), binname(pkg))
 	}
 
-	// if this is a cross compile, add ctxString
-	if pkg.isCrossCompile() {
-		target += "-" + pkg.ctxString()
-	}
-
-	// append build tags
-	if len(pkg.buildtags) > 0 {
-		target += "-" + strings.Join(pkg.buildtags, "-")
-	}
+	target += pkg.Suffix()
 
 	if pkg.gotargetos == "windows" {
 		target += ".exe"
```
(leaving `ctxString` alone since it's used for more things, like `pkg/` subdirectory names where it's important to have `gotargetos-gotargetarch` regardless of cross-compile status)

davecheney:
Fixes #414

### DIFF
--- a/build.go
+++ b/build.go
@@ -317,7 +317,7 @@ func pkgname(pkg *Package) string {
 func binname(pkg *Package) string {
 	switch {
 	case pkg.Scope == "test":
-		return pkg.Name+".test"
+		return pkg.Name + ".test"
 	case pkg.Name == "main":
 		return filepath.Base(filepath.FromSlash(pkg.ImportPath))
 	default:
--- a/build_test.go
+++ b/build_test.go
@@ -20,9 +20,9 @@ func TestBuild(t *testing.T) {
 
 	opts := func(o ...func(*Context) error) []func(*Context) error { return o }
 	tests := []struct {
-		pkg string
+		pkg  string
 		opts []func(*Context) error
-		err error
+		err  error
 	}{{
 		pkg: "a",
 		err: nil,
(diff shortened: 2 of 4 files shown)


tianon:
Nice! :+1:

davecheney:
Nice catch @tianon

tianon:
Heh, I'd already been digging around in that exact code, so it jumped out
at me when I looked at the "-tags" PR. :innocent:

Issue #406, opened 2015-09-22
Compiler output should be relative to a fixed point


davecheney:
Having the compiler output a unique file path when compilation fails has been a long standing issue, see #57, #79, #99.

#57 changed to make paths relative to `$PROJECT`, but this since regressed :(

#395 addressed the regression, by making the output relative to the current working directory.

This is a tracking issue to come up with a better plan than the piecemeal solutions so far.

Requirements:

- [ ] must handle `-R`, ie don't assume that the current working directory is inside the project.
- [ ] must handle temporary files created in `context.Workdir()`, which may not be in the same filesystem namespace as the `$PROJECT`, so may not be compatible with `filepath.Rel`, ie. `C:\TMP` and `D:\SOURCE\PROJECT`. This also includes the generated test packages, and cgo.
- [ ] must be tested, preferably without doing a full end to end test of `cmd/gb`, ie. whatever functions generate the file paths passed to the compiler should be unit tested.

/cc @fatih, @inconshreveable, @tianon

inconshreveable:
I think you mean #377 not #395?

#377 is both unit tested and properly works with context.Workdir() on Windows. (It only modifies paths that were already relative. Any absolute paths are left unchanged.)

Unfortunately, #377's behavior with `-R` is suboptimal. Paths relative to $pwd are probably not the most helpful in that situation. Worse, if you use `-R` and specify a path on a different drive on Windows, gb will now break.

I would suggest one of two alternatives:
1. Paths should always be relative to gb project root
2. Paths should always be relative to $pwd or, if supplied, the argument to `-R`

davecheney:
@inconshreveable correct. To be clear, I am not saying your work on #377 is incorrect. I am grateful for it. This issue is to track the remaining work.

> I would suggest one of two alternatives:
> 1. Paths should always be relative to gb project root
> 2. Paths should always be relative to $pwd or, if supplied, the argument to -R

Sadly I think the solution is going to be more complicated. gb will have to handle both relative and absolute paths. We should be using relative paths internally whereever possible, and probably relative to the project root. But I fear there will be cases where absolute paths are required when dealing with generated code from tests or cgo.

More generally, I am concerned there is too much path munging going on inside build.go and gc.go, which is #99. I think paths should be generated once, and transformed a little as possible after that.

inconshreveable:
Yep, I understand! I just wanted to summarize where #377 leaves the current state of the world for everyone.

[...]

PR #421, opened 2015-10-06
Modify error paths to be relative to the working directory

dradtke:
This commit updates `gb` to automatically modify error paths in a command's output to be relative to the user's working directory, enabling the use of tools like Vim's quickfix to help navigate errors. Before, it would output directory information on individual lines, and the line containing the error would mean nothing to an editor since it would only be the file's base name, but with this patch, `gb` error output is now easily and correctly parsed by editors.

This is a very simple change tailored to my specific use case, and there are probably some changes that could be made to have it integrate better with the rest of `gb`, but it works, and is very straightforward in its functionality.

### DIFF
--- a/gc.go
+++ b/gc.go
@@ -1,10 +1,13 @@
 package gb
 
 import (
+	"bufio"
 	"fmt"
 	"go/build"
+	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 )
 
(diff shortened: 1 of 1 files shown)


davecheney:
Thank you for this proposal. However, I would like to see the solution implemented as per https://github.com/constabulary/gb/issues/406

PR #433, opened 2015-10-30
Add scm option

lateefj:


davecheney:
Thanks for your pukl request. 

Please fix the build errors, and absolutely no relative import paths on this project. Thanks

PR #437, opened 2015-11-03
gb/test: scale test improvements

davecheney:
* gb test now removes the _test/testmain file immediately after executing to
  reclaim space faster.
* build all dependencies immediately, but arrange for linking _test/testmain
  to occur sequentially with execution, so that linking and execution are run
  in the same executor, this prevents multiple testmains being built on disk
  before being run.

### DIFF
--- a/cmd/gb/test.go
+++ b/cmd/gb/test.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"sort"
 
 	"github.com/constabulary/gb"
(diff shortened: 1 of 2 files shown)


davecheney:
With this branch, I can now run all the juju tests, previously it would consume several gb of /tmp (tmpfs) queuing up our enormous test binaries. 

The fact that juju's test binaries are 40mb big is a problem, but that shouldn't stop gb from running the tests.

PR #442, opened 2015-11-08
Always capture command output and report package name

### LINKED ISSUE #440: output failing package before compile errors
Moving the output of the package name til the completion of the compilation, #437 now causes gb to _not_ print the name of the package if there was a compilation failure
```
lucky(~/src/github.com/constabulary/gb) % gb build
gb.go:71: undefined: "github.com/constabulary/gb/log".Fatalf
FATAL command "build" failed: exit status 2
```
It's not clear which package gb.go belongs to. This is because the compile failure happens in an action below the final action in gb.Compile, so the action is skipped, skipping the logging.

This may be ameliorated if #406 is addressed, but it might need a direct fix.

davecheney:
Fixes #440

Prior to #437 the name of the package under compile or test would be
printed *before* the operation started. This was a problem, because the
compilation output could occur later, and not line up with the package
name.

After #437, package name output was surpressed until the compile/test
action completed successfully. This helped improve the output for tests
but as the overall action, the print would never occur, only the output
to stdout from the child process.

This PR solves the problem at the source by delegating the responsibility
for printing the name of the package before any child output. If all
actions complete, then the final action in the set will print the name
of the package.

As it is now unused, `context.run` has been removed.

TODO:

- [ ] decide if the output should go to stderr, not stdout
- [ ] decide if we should include a # prefix as the go tool does

### DIFF
--- a/cgo.go
+++ b/cgo.go
@@ -3,12 +3,15 @@ package gb
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/constabulary/gb/log"
 )
 
 func cgo(pkg *Package) (*Action, []string, []string, error) {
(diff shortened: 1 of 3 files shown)


davecheney:
I've tested this change today and I'm going to commit this as is without addressing the TODO items.

I'll address the TODO items once 0.3.1 is cut.

PR #449, opened 2015-11-11
Show vendor plugin help when no command is given

### LINKED ISSUE #446: gb vendor error
$ gb vendor
FATAL: command "vendor" failed: plugin: no command supplied

Go 1.5.1

kalafut:
I don't have much familiarity with gb internals, but I use it a lot and saw the "Help Wanted" sign.  This PR may not be exactly what you're looking for. I couldn't come up with a good reason why args < 1 should cause an error at the gb level... why not let the plugins worry about that?  Also, the various exit codes in gb-vendor were causing ugly "FATAL: ..." at the end of the help, so I pulled them.

Fixes #446

### DIFF
--- a/cmd/gb-vendor/main.go
+++ b/cmd/gb-vendor/main.go
@@ -18,10 +18,7 @@ var (
 )
 
 func init() {
-	fs.Usage = func() {
-		printUsage(os.Stderr)
-		os.Exit(2)
-	}
+	fs.Usage = usage
 }
 
 var commands = []*cmd.Command{
@@ -43,8 +40,8 @@ func main() {
 
 	switch {
 	case len(args) < 1, args[0] == "-h", args[0] == "-help":
-		fs.Usage()
-		os.Exit(1)
+		printUsage(os.Stdout)
+		os.Exit(0)
 	case args[0] == "help":
 		help(args[1:])
 		return
--- a/cmd/gb/main.go
+++ b/cmd/gb/main.go
@@ -63,9 +63,6 @@ func main() {
 		}
 		command = &cmd.Command{
 			Run: func(ctx *gb.Context, args []string) error {
-				if len(args) < 1 {
-					return fmt.Errorf("plugin: no command supplied")
-				}
 				args = append([]string{plugin}, args...)
 
 				env := cmd.MergeEnv(os.Environ(), map[string]string{


davecheney:
Thanks for this, but I don't think this is quiet right. Looking at `cmd/gb` it does something like this
```
        args := os.Args
        if len(args) < 2 || args[1] == "-h" {
                fs.Usage()
                os.Exit(1)
        }
```
If you adapt this to gb-vendor and use os.Exit(0) with a comment explaining what this does to the caller. This should be all that is needed.

kalafut:
Hi Dave,

I'll revisit this, though I do have a couple of questions:

1. What do you want fs.Usage() to do in general with respect to error codes?  In your code snippet, fs.Usage() (set to usage() in cmd/gb/help.go) prints the usage and then exits with error code 2. The os.Exit(1) is never hit. This also results in: 

    ```
    gb help    # (prints help, exit code 0, which seems correct BTW)
    gb         # (prints identical help, exit code 2)
    ```

    Within gb-vendor there is an unused usage() function, and then fs.Usage defined in init().  Looks     like this could be consolidated but I'd like to know your goal for the help system.

2. I wasn't sure if your last sentence indicated that `cmd/gb/main.go` didn't need to change. I think it has to, but I wanted to double check..

Thanks,
Jim

davecheney:
>
>
>    1.
>
>    ]cmd/gb/help.go) prints the usage and then exits with error code 2.
>    The os.Exit(1) is never hit. This also results in:
>
> Oh, I didn't realise that.


>
>    1.
>
>    gb help    # (prints help, exit code 0, which seems correct BTW)
>    gb         # (prints identical help, exit code 2)
>
>    Within gb-vendor there is an unused usage() function, and then
>    fs.Usage defined in init(). Looks like this could be consolidated but I'd
>    like to know your goal for the help system.
>
> I don't have much of a goal for the help system, it's pretty minimal at
the moment, idiosyncratic, and slightly broken, sadly it's not top of the
list atm.


>
>    1.
>    2.
>
>    I wasn't sure if your last sentence indicated that cmd/gb/main.go
>    didn't need to change. I think it has to, but I wanted to double check..
>
> It's fine for gb to exit with error 2, that is a sign to the caller they
got something wrong (possibly scripted), gb-vendor isn't intended to be
called directly, so if we need to make the exit codes change to suppress a
message from gb (which thinks it is calling a plugin), that is fine.

kalafut:
I thought about it some more and since the custom fs.Usage() function is called by FlagSet when parsing bad input, keeping the os.Exit() there makes sense.

I updated the PR to just call printUsage() directly in case of default help and then exit 0. This provides clean behavior for `gb vendor` without any error text.

davecheney:
LGTM. Thank you very much.

Don't worry about the travis failure, we're just getting it going and haven't cut over yet.

PR #454, opened 2015-11-16
vendor: reactivate bitbucket tests

### LINKED ISSUE #452: gb/vendor: fix bitbucket tests
#451 disabled the bitbucket tests in vendor/repo_test.go as some interaction with mercurial in a child process was trying to do something interactive and hanging.

davecheney:
Fixes #452

Because of reasons bitbucket.org is not in the ssh known hosts' file in tavis.yml. This leaves two questions

1. why does it appear to be in the ssh hosts' file on the other CI providers
2. why, even though we assign /dev/null to Stdin, does mercurial still prompt for an answer

Having no good answer for both, try to work around the problem by pre-populating the known hosts file during setup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ install:
   - echo '#!/bin/bash' > "$GOPATH/bin/sudo"
     && echo 'echo >&2 attempted sudo "$@"' >> "$GOPATH/bin/sudo"
     && chmod +x "$GOPATH/bin/sudo"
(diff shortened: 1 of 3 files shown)


tianon:
Hahaha, this is pretty similar to the workaround I was contemplating.  Dirty indeed, but super effective. :+1:

davecheney:
@tianon when in doubt, sell out.

solarce:
there is also http://docs.travis-ci.com/user/ssh-known-hosts/ :)

PR #455, opened 2015-11-18
Improve the fix for bitbucket.org hosts

davecheney:
Thanks @solarce


tianon:
Omg that's a shiny feature! :heart_eyes:

PR #456, opened 2015-11-19
Added SIGINT handling in context

### LINKED ISSUE #436: gb should catch ^C and issue ctx.Destroy

davidkaya:
Fixes #436 

Not quite sure about the return code though.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/signal"
 	"path/filepath"
 
 	"github.com/constabulary/gb"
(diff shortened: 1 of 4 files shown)


davecheney:
Thanks for taking a look at this. I would like to see the logic move to the command, not context.

davidkaya:
@davecheney Dave? I tried calling it in RunCommand, but it seems it does not get called. Do you know where the problem could be?

davecheney:
@davidkaya this code should not go in gb.Context, there may be more than one context in use per program.

davidkaya:
Yep, I know, but I have tried to put the sigint handling into RunCommand in cmd.go and it does not seem to get called.

davidkaya:
@davecheney I've moved it to command.

davecheney:
I'm really sorry to say this but this is not what I want. I'm very sorry, because I know you are volunteering your time and effort for free. Can we please move this discussion to the issue to talk about the implementation, then do the implementation.

Thanks

davidkaya:
@davecheney Sure, no problem :)

davecheney:
Thanks. FWIW I think each main should have something like

var onExit []func() error

func exit(code int) {
    for _, fn := range onExit {
          fn()
    }
    os.Exit(code)
}

Then replace the usual defers with appends to onExit.

davecheney:
Also, cmd/gb needs to ignore ^C, not hook and exit. See the way it's done in cmd/go

davidkaya:
I will create new PR for the signal handling.

PR #458, opened 2015-11-22
Added system signal handling

### LINKED ISSUE #436: gb should catch ^C and issue ctx.Destroy

davidkaya:
Fixes #436 

Not ready for merge - cmd/gb-vendor is not implemented yet.

### DIFF
--- a/signal.go
+++ b/signal.go
@@ -1,0 +1,26 @@
+package gb
+
+import (
+	"os"
+	"os/signal"
+	"sync"
+)
+
+// If we catch signal then this will get closed
+var interrupted = make(chan struct{})
+
+func processSignal() {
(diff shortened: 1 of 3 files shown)


davecheney:
I'm really sorry but this is not right. I know this is the third time I have rejected your proposal. Would you let me take a go tomorrow and lay out the design that I would like to see ?

signal.go
```diff
@@ -0,0 +1,26 @@
+package gb
```
davecheney: This code should not live in the gb package, it should know nothing about it. Ideally this would live in the cmd or cmd/gb package.

davidkaya:
Yep, sure :) no problem.

Issue #445, opened 2015-11-09
^C during compile leaves broken binaries


davecheney:
```
lucky(~/src/github.com/constabulary/gb) % gb build
github.com/constabulary/gb/log
github.com/constabulary/gb/vendor
github.com/constabulary/gb
github.com/constabulary/gb/cmd
github.com/constabulary/gb/test
^Clucky(~/src/github.com/constabulary/gb) % gb build
-bash: /home/dfc/bin/gb: cannot execute binary file: Exec format error
```
This is because `gb.Link` links directly to the final binary path, rather than $WORKDIR and then a copyfile step.

davecheney:
For extra sadface, the go tool is too cautious to overwrite the dodgy blob that was half written
```
lucky(~/src/github.com/constabulary/gb) % go install -v ./...
github.com/constabulary/gb/log
github.com/constabulary/gb/vendor
github.com/constabulary/gb
github.com/constabulary/gb/cmd
github.com/constabulary/gb/test
github.com/constabulary/gb/cmd/gb-vendor
github.com/constabulary/gb/cmd/gb
go install github.com/constabulary/gb/cmd/gb-vendor: build output "/home/dfc/bin/gb-vendor" already exists and is not an object file
go install github.com/constabulary/gb/cmd/gb: build output "/home/dfc/bin/gb" already exists and is not an object file
```

PR #459, opened 2015-11-23
cmd/gb: catch signals during build and test

### LINKED ISSUE #436: gb should catch ^C and issue ctx.Destroy

davecheney:
This is the logic that Alex Brainman and I added to cmd/go back in Go 1.0. It's not perfect, but it's worked well enough for the last few years.
 
### Todo
- [x] gb.Executors need to take some kind of reference to the interrupted channel so it can shutdown
- [ ] No tests for interruption yet

Fixes #436
Updates #445

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -1,6 +1,7 @@
 package gb
 
 import (
+	"errors"
 	"sync"
 )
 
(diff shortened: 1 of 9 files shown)


davidkaya:
Oh, so I was not that far from the correct implementation. I just placed it into wrong package :) Should I add the logic to the cmd/gb-vendor also (or atleast give it a try ;))?

Thanks for explenation! :)

davecheney:
@davidkaya this is a bit rough, but ready for review if you have time.

executor.go
```diff
@@ -83,9 +84,14 @@ func ExecuteConcurrent(a *Action, n int) error {
 				}
 			}
 			// wait for a permit and execute our action
-			<-permits
-			result <- a.Run()
-			permits <- true
+			select {
```
davidkaya: Isn't select non-blocking?

davecheney: Select is only non blocking if there is a default: clause.

davidkaya: Oh, ok, good to know :)

davidkaya:
LGTM. :)

PR #464, opened 2015-11-26
build: add support for .syso files

nbjahan:
this is based on: 
[golang/go/src/cmd/go/build.go#L1487-L1490](https://github.com/golang/go/blob/21efa7b2bc872958bcb252f5ab4dc52b2b0abeae/src/cmd/go/build.go#L1487-L1490)
[golang/go/src/cmd/go/build.go#L3056](https://github.com/golang/go/blob/21efa7b2bc872958bcb252f5ab4dc52b2b0abeae/src/cmd/go/build.go#L3056)

My use case was embedding the resource file inside cross compiled .exe without cgo
using: rsrc tool: [akavel/rsrc](https://github.com/akavel/rsrc)
or windres: lxn/walk#108
and it worked, I'm not sure how to write the test for this or if the cgo related changes are correct. 
Please advice, Thanks.

### DIFF
--- a/build.go
+++ b/build.go
@@ -155,6 +155,11 @@ func Compile(pkg *Package, deps ...*Action) (*Action, error) {
 		ofiles = append(ofiles, ofile)
 	}
 
+	// step 4. add system object files.
+	for _, syso := range pkg.SysoFiles {
+		ofiles = append(ofiles, filepath.Join(pkg.Dir, syso))
+	}
+
 	build := &compile
 
 	// Do we need to pack ? Yes, replace build action with pack.
--- a/cgo.go
+++ b/cgo.go
@@ -69,6 +69,7 @@ func cgo14(pkg *Package) (*Action, []string, []string, error) {
 	cflags := append(cgoCPPFLAGS, cgoCFLAGS...)
 	cxxflags := append(cgoCPPFLAGS, cgoCXXFLAGS...)
 	gcc1, ofiles := cgocc(pkg, cflags, cxxflags, cfiles, pkg.CXXFiles, runcgo1...)
+	ofiles = append(ofiles, pkg.SysoFiles...)
 	ofile := filepath.Join(filepath.Dir(ofiles[0]), "_cgo_.o")
 	gcc2 := Action{
 		Name: "gccld: " + pkg.ImportPath + ": _cgo_.o",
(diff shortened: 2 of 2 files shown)


davecheney:
LGTM. There is not test case, do you have a project that I can add to the integration test suite to make sure I don't break this in the future ?

nbjahan:
Thanks Dave,

Here is my attempt to test this: https://github.com/nbjahan/gb-sysofiles-test

__`src/nocgo`__
there is a `rsrc.syso` in `src/nocgo` which contains the `main.manifest`

we build a binary with: 
```bash
GOOS=windows GOARCH=386 gb build -f nocgo
```
and test it by checking if the .syso file is included in the binary:
 ```bash
strings bin/nocgo-windows-386.exe |grep -q SomeFunkyNameHere && echo "Passed" || echo "Not Passed"
```

__`src/cgo`__
the same `.syso` file is at `src/cgo` but the loader complains about the file architecture:
```bash
$ gb build -f cgo
ld: warning: ignoring file rsrs.syso, file was built for unsupported file format ( 0x4C 0x01 0x01 0x00 0x00 0x00 0x00 0x00 0x6F 0x03 0x00 0x00 0x01 0x00 0x00 0x00 ) which is not the architecture being linked (x86_64): rsrs.syso
```
maybe its related to: https://github.com/golang/go/wiki/GcToolchainTricks
> Drawback of this issue is that it seems incompatible to cgo, so only use it when you don't use cgo, at least for now. I (minux) is working on figuring out why they're incompatible.

so I test cgo buy looking for `'ld: warning: ignoring file rsrs.syso'` this way we know that gb tried to pack `rsrc.syso` file.

I couldn't find a way to access the resource data in darwin/linux in go1.4+

By the way I found the commit on go repo: https://github.com/golang/go/commit/b0996334c1cd2c14e07fb4f17c924cdf698ae48d

davecheney:
I'm sorry, I wasn't very clear about what I meant by an integration test.

https://github.com/constabulary/integration-tests

Is an extra repo of gb like projects where the build job will try to build
some (not all, some are disabled) of the projects after a unit test. A way
to test this .syso addition would be to add an extra gb project to this
repo and we'll arrange to run it as part of the CI job.

Issue #473, opened 2015-12-06
gb test: Failed to resolve import path "-v"


nsf:
I don't know if this is expected. But it is confusing. I know there is some logic to recognize known testing flags and work with them in a particular way, but I find this confusing.

Define a "projectx" package with TestX and TestY functions in `main_test.go` file.

`gb test -v projectx -run TestX` - runs TestX, all fine
`gb test -v -run TestX projectx` - runs TestX, all fine

It makes you think that position of flags relative to package name is not important. So you try this:

`gb test -v projectx` - runs all tests, all fine
`gb test projectx -v` - fails with:

FATAL: command "test" failed: failed to resolve import path "-v": cannot find package "-v" in any of

I tried to look at the code and figure out what makes "-v" so special. Nothing, other than it's defined on the top level command here: https://github.com/constabulary/gb/blob/master/cmd/gb/test.go#L33

And it is indeed related, because `-cover` fails the same way. Even though both of them are defined here: https://github.com/constabulary/gb/blob/master/cmd/gb/testflag.go#L23-L38

Also if we take different boolean flags which are defined in that list, but are not present in top level package definition, they work just fine. For example `-benchmem` flag can be specified anywhere related to package name.

davecheney:
Thanks for raising this issue. I think what is happening here is gb's flag
parsing is stopping at projectx, and the reminder of the flags are being
forwarded to the test binary which reject them as -v is not defined there.

At a guess I think this is expected, but I'll confirm and see what the go
tool does.

Issue #476, opened 2015-12-08
gb env should return value associated with a variable 


gosuri:
A cool enhancement that'll make `gb env` more script friendly would be to return the assigned value for a var, similar to `go env VAR`.  Example:
```sh
$ gb env GB_PROJECT_DIR
/Users/gosuri/code/walker
```

 I can send a patch, if you'd like.

davecheney:
:+1: from me. You'll probably have to do some serious surgery to replace the simplistic way `gb info` works now. 

Alternatively you may want to add a `gb env`, or make `gb env` an alias for `gb info`. 
 
Either way, please make sure you add end to end tests to cmd/gb/gb_test.go.

Thanks

Dave

gosuri:
Awesome! Will get rolling.

gosuri:
I see what you meant by "serious surgery", the args are pkg paths :)

davecheney:
I believe there is an option to override arg parsing per command. Have a
look at the cmd.Cmd structure and check out how test monkeys with argument
parsing.

gosuri:
Thanks.. Figured it out

davecheney:
Sorry its such a hand rolled mess.

Part of the challenge with gb is I've made such a song and dance about how
package management should be done, I've sort of painted myself into a
corner and made myself unable to use third party packages.

I should probably find a ladder and get down off my high horse.

gosuri:
I sort of jumped straight to the command's code under the assumption that args weren't manipulated. It wasn't hard tracing back. The code and tests were clear and easy to understand.

More over, I learned much more about the tool with this exercise. Will come handy since I'll be using gb everyday.

davecheney:
closed by #477

PR #477, opened 2015-12-08
cmd/gb: info and env returns values for env vars

gosuri:
`gb info` optionally accepts a list of environment variable names and
returns their values. `gb env` is aliased to gb info.

Updates #476

### DIFF
--- a/cmd/gb/env.go
+++ b/cmd/gb/env.go
@@ -19,6 +19,8 @@ given as arguments, env prints the value of each named variable on its own line.
 `,
 	Run: info,
 	ParseArgs: func(ctx *gb.Context, cwd string, args []string) []string {
+		// env treats arguments as environment variables names,
+		// don't do any processing.
 		return args
 	},
 }
--- a/cmd/gb/env.go
+++ b/cmd/gb/env.go
@@ -15,7 +15,7 @@ var EnvCmd = &cmd.Command{
 	Short:     "print project environment variables",
 	Long: `
 Env prints project environment variables. If one or more variable names is 
-given as arguments, info prints the value of each named variable on its own line.
+given as arguments, env prints the value of each named variable on its own line.
 `,
 	Run: info,
 	ParseArgs: func(ctx *gb.Context, cwd string, args []string) []string {
--- a/cmd/gb/env.go
+++ b/cmd/gb/env.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"fmt"
-
 	"github.com/constabulary/gb"
 	"github.com/constabulary/gb/cmd"
 )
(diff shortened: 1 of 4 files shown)


cmd/gb/env.go
```diff
@@ -13,37 +11,14 @@ func init() {
 
 var EnvCmd = &cmd.Command{
 	Name:      "env",
-	UsageLine: `env`,
+	UsageLine: `env [var ...]`,
```
davecheney: nice. I'd really like to use a getopt style package to automate a lot of this.

cmd/gb/env.go
```diff
@@ -13,37 +11,14 @@ func init() {
...
-	return ""
-}
-
-func makeenv(ctx *gb.Context) []envvar {
-	return []envvar{
-		{"GB_PROJECT_DIR", ctx.Projectdir()},
-	}
+	Run: info,
+	ParseArgs: func(ctx *gb.Context, cwd string, args []string) []string {
+		return args
```
davecheney: a comment here would be nice

// env treats arguments as environment variables names, don't do any processing.

cmd/gb/gb_test.go
```diff
@@ -589,6 +589,17 @@ func TestInfoCmd(t *testing.T) {
 	gb.grepStdout(`^GB_GOROOT="`+regexp.QuoteMeta(runtime.GOROOT())+`"$`, "missing GB_GOROOT")
 }
 
+func TestInfoWithArgs(t *testing.T) {
+	gb := T{T: t}
+	defer gb.cleanup()
+
+	gb.tempDir("src")
+	gb.cd(gb.tempdir)
+	gb.run("info", "GB_PROJECT_DIR", "GB_GOROOT")
```
davecheney: could you add a test for missing something like 

`gb info GB_PROJECT_DIR GB_MISSING"

Should return one line of output and _not_ complain about GB_MISSING.

cmd/gb/info.go
```diff
@@ -32,22 +32,27 @@ Values:
...
-	vars := []struct{ name, value string }{
-		{"GB_PROJECT_DIR", ctx.Projectdir()},
-		{"GB_SRC_PATH", joinlist(ctx.Srcdirs()...)},
-		{"GB_PKG_DIR", ctx.Pkgdir()},
-		{"GB_BIN_SUFFIX", ctx.Suffix()},
-		{"GB_GOROOT", runtime.GOROOT()},
+	env := makeenv(ctx)
+	if len(args) > 0 {
+		for _, arg := range args {
+			fmt.Println(findenv(env, arg))
```
davecheney: I think there should be a check here that if findenv returns "", skip the output.

davecheney:
Looks pretty good to me, thank you. Can you please address these minor items. Then we need to update  alldoc.go which is driven by go generate. To do this

go install .../cmd/gb
cd .../cmd/gb
go generate

will update alldocs.go.

cmd/gb/env.go
gosuri: Thank you. I cound’t find a decent way to parse arguments (only flags) using getopt. Implemented it simlar to the https://github.com/golang/go/blob/master/src/cmd/go/env.go#L16

cmd/gb/env.go
gosuri: :+1:

cmd/gb/gb_test.go
gosuri: It returns a blank line for each invalid variable, for API simplicity (similar to `go env`).

Example:
```sh
$ gb env GB_PROJECT_DIR GB_UNKNOWN GB_GOROOT
/Users/gosuri/code/go

/usr/local/Cellar/go/1.5.2/libexec
```

cmd/gb/info.go
gosuri: I use the "" to render a blank line, not checking on purpose.

davecheney:
Lovely, thanks very much. This will go out in the 0.3.4 release once I've fixed #473

gosuri:
Sweet! Thank you

PR #479, opened 2015-12-11
manifest: ensure nul termination

thorduri:
Prevents creating a diff when opening the file in an editor and compulsively writing before exiting...


davecheney:
not lgtm, sorry. Can you please raise an issue, discuss your change, then implement your change. There may be a better or simpler way to achieve what you need.

thorduri:
Not worth it, ty tho.

PR #488, opened 2015-12-15
Fix `gb vendor fetch goji.io`

nullstyle:
Prior to this PR, attempting to run `gb vendor fetch goji.io` would error out with a "not a valid import path" error.  This was unexpected since `go get goji.io` works fine.


davecheney:
If this pattern is supported by go get then sgtm. Thanks for this change. 

Please fix the build failure.

nullstyle:
My bad!  Things should be green now.

davecheney:
LGTM. Thanks.

PR #499, opened 2015-12-17
Refactor goversion support

davecheney:
Previously we had two constants, go14 and go15, but adding a third, go16 makes
it clear that the design is wrong.

Replace this with one constant, `goversion` which can be compared with other
constants literals at compile time. eg, `if goversion > 1.4 { // go 1.5 and later }`.

### DIFF
--- a/cgo.go
+++ b/cgo.go
@@ -14,9 +14,9 @@ import (
 
 func cgo(pkg *Package) (*Action, []string, []string, error) {
 	switch {
-	case gc14:
+	case goversion == 1.4:
 		return cgo14(pkg)
-	case gc15:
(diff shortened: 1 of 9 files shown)


tianon:
If/when Go gets to 1.10, won't this have to change again?

davecheney:
Probably, I'm prepared to take that as TODO for the 2017/18 timeframe.

tianon:
:+1:

davecheney:
TBH, this is only necessary because I want to treat Go 1.6 differently to Go 1.5. If this wasn't necessary, I wouldn't need to keep adding magic files.

PR #520, opened 2015-12-30
Change `go` binary file lookup for generate & test tools

mguzelevich:
`filepath.Join(runtime.GOROOT(), "bin", "go")` is not always find the correct location of a `go` binary file

ex.:
- nix os

```
runtime.GOROOT=/nix/store/c4vlrc0bfl5nv85ph9p0hhz2a5h59lkm-go-1.5.1/share/go

% whereis go
go: /nix/store/ys47s890vyrf1vghzhf3vsnnl8k1m6gg-system-path/bin/go

% ls -lah /nix/store/ys47s890vyrf1vghzhf3vsnnl8k1m6gg-system-path/bin/go
/nix/store/ys47s890vyrf1vghzhf3vsnnl8k1m6gg-system-path/bin/go -> /nix/store/c4vlrc0bfl5nv85ph9p0hhz2a5h59lkm-go-1.5.1/bin/go
```
`/nix/store/c4vlrc0bfl5nv85ph9p0hhz2a5h59lkm-go-1.5.1/share/go/bin/go` - is not correct location

### DIFF
--- a/cmd/gb/gb_test.go
+++ b/cmd/gb/gb_test.go
@@ -65,7 +65,14 @@ func TestMain(m *testing.M) {
 			os.Exit(2)
 		}
 		testgb = filepath.Join(dir, testgb)
-		out, err := exec.Command(filepath.Join(runtime.GOROOT(), "bin", "go"), "build", "-o", testgb).CombinedOutput()
+
+		goBinary, err := lookupGo()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, err)
+			os.Exit(1)
+		}
+
+		out, err := exec.Command(goBinary, "build", "-o", testgb).CombinedOutput()
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "building testgb failed: %v\n%s", err, out)
 			os.Exit(2)
--- a/cmd/gb/generate.go
+++ b/cmd/gb/generate.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"runtime"
 
 	"github.com/constabulary/gb"
 	"github.com/constabulary/gb/cmd"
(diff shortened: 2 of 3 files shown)


davecheney:
Not LGTM sorry. We cannot use the go.binary in the path as this may not be the same go binary that built this go test binary.

davecheney:
Please log an issue to discuss the problem with nix and we'll workshop a solution there.

mguzelevich:
ubuntu 15.04 docker image via AWS
```
FROM ubuntu:15.04
...
RUN apt-get update && apt-get install -y --no-install-recommends software-properties-common
RUN add-apt-repository ppa:evarlast/golang1.5
RUN apt-get update && apt-get install -y --no-install-recommends golang
...
ENV GOPATH $home/gopath
...
RUN go get github.com/constabulary/gb/...
...
```

same problem.
`gb generate` - failed with `FATAL: command "generate" failed: fork/exec /usr/lib/go/bin/go: no such file or directory`

```
# uname -a
Linux 12d6e321aac2 3.14.44-32.39.amzn1.x86_64 #1 SMP Thu Jun 11 20:33:38 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

# go env | grep GOROOT
GOROOT="/usr/lib/go"

# ls -la /usr/lib/go
total 16
drwxr-xr-x  4 root root 4096 Dec 28 16:35 .
drwxr-xr-x 35 root root 4096 Dec 28 16:37 ..
lrwxrwxrwx  1 root root   31 Aug 21 18:16 doc -> ../../share/doc/golang-doc/html
lrwxrwxrwx  1 root root   38 Aug 21 18:16 favicon.ico -> ../../share/doc/golang-doc/favicon.ico
drwxr-xr-x  2 root root 4096 Dec 28 16:35 lib
drwxr-xr-x  6 root root 4096 Aug 21 18:07 pkg
lrwxrwxrwx  1 root root   18 Aug 21 18:16 src -> ../../share/go/src
```

mguzelevich:
`gb doc` - failed with `FATAL: command "generate" failed: fork/exec /usr/lib/go/bin/godoc: no such file or directory`

davecheney:
Thanks for the report. I think this is fixable, it looks like Ubuntu move
the go tool relative to GOROOT,  think if we try GOROOT, GOROOT/.. And
GOROOT/../.. That should work.

davecheney:
Go gb generate I think it would be easier to just execute "go" and let the
normal path lookup do its magic

davecheney:
For the cmd/gb test it is important to call the save version of Go that
built the test binary. If that cannot be derived from GOROOT, we'll have to
skip the whole test.

PR #524, opened 2015-12-30
cmd/gb: handle various go installation permutations

### LINKED ISSUE #521: gb generate / gb doc does not work with packaged version of Go
gb doc - failed with FATAL: command "generate" failed: fork/exec /usr/lib/go/bin/godoc: no such file or directory

gb doc, gb generate and go test cmd/gb does not work when using a packaged version of Go if that version has spread GOROOT out into several components.

For gb doc and gb generate, just call "go" and let LookPath figured it out. For go test cmd/gb we must use the same GOROOT that built the test binary.

See #520

davecheney:
Fixes #521

Try to locate `bin/go` in the various permutations that package managers use.

### DIFF
--- a/cmd/gb/gb_test.go
+++ b/cmd/gb/gb_test.go
@@ -65,12 +65,23 @@ func TestMain(m *testing.M) {
 			os.Exit(2)
 		}
 		testgb = filepath.Join(dir, testgb)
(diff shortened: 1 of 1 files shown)


davecheney:
@mguzelevich can you please try this.

PR #185, opened 2015-06-03
cmd/gb-vendor: move copypath to vendor.Copypath, skip symlinks

davecheney:
Fixes #184 

Ignore symlinks in the src directory.

Also mutes debugging output from `copyfile`

Issue #184, opened 2015-06-03
cmd/gb-vendor: copyfile does not handle invalid symlinks


davecheney:
`copyfile` pukes if it hits an invalid symlink.

Go does not support symlinks inside the source directory, and this example is from a test fixture, and we don't run tests inside `$PROJECT/vendor`. The simplest approach would be to skip symlinks.
```
copyfile(dst: /home/dfc/devel/juju/vendor/src/gopkg.in/juju/charm.v5/internal/test-charm-repo/series/format2/build/ignored, src: /tmp/gb-vendor-720588324/internal/test-charm-repo/series/format2/build/ignored)
FATAL command "fetch" failed: copyfile: open("/tmp/gb-vendor-720588324/internal/test-charm-repo/series/format2/hooks/symlink"): open /tmp/gb-vendor-720588324/internal/test-charm-repo/series/format2/hooks/symlink: no such file or directory
FATAL command "vendor" failed: exit status 1
```

PR #534, opened 2016-01-09
fileutils: add support for copying symlinks to Copypath

FiloSottile:
This was removed instead of being special cased in #185, because gb has no
direct need for it (see #184).  However it's useful for other users of the
library.  For example gvt needs to retain test fixtures intact (as vendor/
packages are testable) and needs to support symlink-based packages (since
the go build tool does).

Fixes FiloSottile/gvt#12

### DIFF
--- a/fileutils/fileutils.go
+++ b/fileutils/fileutils.go
@@ -32,14 +32,12 @@ func Copypath(dst string, src string) error {
 			return nil
 		}
 
+		dst := filepath.Join(dst, path[len(src):])
+
 		if info.Mode()&os.ModeSymlink != 0 {
-			if debugCopypath {
-				fmt.Printf("skipping symlink: %v\n", path)
-			}
-			return nil
+			return Copylink(dst, path)
 		}
 
-		dst := filepath.Join(dst, path[len(src):])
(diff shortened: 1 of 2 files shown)


davecheney:
@FiloSottile Thanks for adding this. I haven't been clear about the direction of this package at all, but the intention is to move it out to another package, probably github.com/pkg/fileutils, soon.

Instead of carrying this feature inside gb where i'll probably drop it later and break you again, why don't I make the move of fileutils now, and then that package can evolve on it's own ?

How does that sound ?

FiloSottile:
Sure, sounds great! I am vendoring fileutils like a normal package in gvt anyway, so I can switch easily.

davecheney:
Just a FYI, I'm working on the replacement for this package in github.com/pkg/fileutils. It isn't quite ready for prime time yet, but shouldn't take much longer.

FiloSottile:
👍

As soon as you switch to that on gb, let me know and I'll port gvt

FiloSottile:
Should I port this to github.com/pkg/fileutils?

davecheney:
Probably, but wait a bit, i'm behind on finishing that package.

PR #535, opened 2016-01-09
vendor: allow ~ in import path

FiloSottile:
~ is an allowed character in the import path, and it's used in the wild for user repositories


davecheney:
This looks ok, but I'd like to add a test, preferably in vendor/repo_test.go (from memory) to check that this works. Do you have a url that has a ~ in it that I can add as a test case ?

FiloSottile:
Makes sense, but sadly I don't have a valid public one.

davecheney:
That's ok. We'll just have to live with it if I break it in the future.

Issue #536, opened 2016-01-09
[rfc] gb build/test automatically download missing dependencies


davecheney:
This rather blandly titled issue is a placeholder to discuss the changes to gb for the 0.4 series, which the roadmap hints _"improve gb vendor"_.

# Backstory
gb cares about reliable builds, a lot. Giving Go developers the tools they need to achieve repeatable builds was the motivation for developing gb, and the main way gb does this is via the `$PROJECT/vendor/src` directory. A bit later `gb-vendor` came along when it was clear that users wanted tooling to help them manage their vendor'd code.

But it also clear that not everyone is comfortable with actually having copies of the source in their tree; they would rather have a manifest file that explains where to get those dependencies on request. This catalyzed around the `gb vendor restore` command which I now regret accepting as it confused the message that gb's reliable build story is based on vendoring. See #498.

To be very clear, the answer for how to get the most reliable builds with gb is always to copy your dependencies into `$PROJECT/vendor/src`. But sometimes there is an argument for trading reliability for convenience, this is the motivation for this proposal.

# Proposal
I proposal that during package resolution (which happens for build and test, but may happen at other stages in the future) once the standard search locations have been exhausted (see below), a per user cache of source code is consulted, and if the package is not found in the cache, we attempt to fetch it using the standard `gb vendor fetch` lookup rules. 

Importantly, this is not a proposal to automatically invoke `gb vendor fetch`, but a separate mechanism that is invoked once all the _per project_ source locations have been exhausted.

TODO:
- [ ] decide if this is opt in or opt out; should there be a flag to disable / enable this behaviour, and if so, what its default state will be?

# Updating
Maintaining a cache means establishing rules for when the cache is updated. My proposal for this is updating the cache, ie fetching from upstream if the entry _is_ present in the cache, should be off by default.

This means adding a new flag, probably `-u` which effectively marks anything in the cache as stale and forces the upstream to be consulted. This could be a potentially very time consuming operation.

TODO:
- [ ] decide if a partial update is something that should be supported. The UI for this sounds complicated, so it might be simpler to give the user a command to clear a particular cached entry, which accomplishes the same thing.

# Package cache
The package cache, or more correctly a cache of source code, is a location where dependencies can be stored and referenced for gb projects. This cache is _per user_ not _per project_.

TODO:
- [ ] figure out the appropriate location for the package cache, some operating systems have a notion of a per user cache, others don't.

# Search order
The search order for packages is currently:

- `$GOROOT`
[...]

PR #537, opened 2016-01-12
new package: depfile

davecheney:
This package parses an io.Reader consisting of key/value pairs into a `map[string]map[string]string`. The format of the line is

    name key=value [key=value]...

The intention is this package will be used to parse a dependency file, or `depfile`, consisting of lines like this
```
github.com/pkg/profile version=0.1.0
github.com/docker/docker version=1.7.0
```
_note:_ this package does not assign semantic meaning to the contents of the `map[string]map[string]string`.

### DIFF
--- a/depfile/depfile.go
+++ b/depfile/depfile.go
@@ -1,0 +1,153 @@
+// depfile loads a file of tagged key value pairs.
+package depfile
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+)
+
+// ParseFile parses path into a tagged key value map.
+// See Parse for the syntax of the file.
+func ParseFile(path string) (map[string]map[string]string, error) {
+	r, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("%s:%v", path, err)
+	}
+	defer r.Close()
+	return Parse(r)
+}
+
+// Parse parses the contents of r into a tagged key value map.
+// If successful Parse returns a map[string]map[string]string.
+// The format of the line is
+//
+//     name key=value [key=value]...
+//
+// Elements can be seperated by whitespace (space and tab).
+// Lines that do not begin with a letter or number are ignored. This
+// provides a simple mechanism for commentary
+//
+//     # some comment
+//     github.com/pkg/profile version=0.1.0
+//
+//     ; some other comment
+//     // third kind of comment
+//       lines starting with blank lines are also ignored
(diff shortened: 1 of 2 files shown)


tianon:
This is a really simple file format! Easy to read, write, and parse, even in raw shell scripting. +1 here

tianon:
Do you envision specifying version ranges or constraints in this format somehow later, or planning to keep it really simple?

kostya-sh:
https://xkcd.com/927/ ;)

davecheney:
@tianon 

> Do you envision specifying version ranges or constraints in this format somehow later, or planning to keep it really simple?

This package just parses the file into `map[string]map[string]string`, it's up to the caller to interpret what that means.

wrt. questions about ranges and constraints, the decision in #536 is to use this file to specify a direct mapping of import path prefix to version number. Maybe if gb is extended to work with libraries, ranges could be added, but that's up to the calling code to interpret it, not this package.

davecheney:
@kostya-sh yup, that's about the size of it. I'm tired of JSON's inability to deal with numbers, and there are no other encodings save text/csv and xml in the standard library, so this will suffice.

tianon:
Right, makes sense -- I was just curious if you think this format is
expressive enough to cover that use case (which is likely to come up
eventually). :+1:

davecheney:
@tianon should be, it's just `key=value`, you can have multiple kv pairs. Here is an example of an valid, but entirely bogus depfile 
```
// use docker 1.7.0
github.com/docker/docker version=1.7.0 vcs=git

// use paultag's fork
github.com/docker/machine remote=https://github.com/paultag/machine rev=ebcidic
```

tianon:
Yeah, apologies -- I didn't want to get this straying too far from the actual file format (since how to use the file format for dependencies is definitely a separate discussion), but IMO thinking about the use case for the format is helpful when designing it. :+1:

Seems sane enough to me!

davecheney:
One case which is not covered is quoting whitespace. I don't plan to add it til it is needed.

kostya-sh:
@davecheney, I am actually quite like this format. Still I beleive having a single format for all Go dependency management tools would be better.

Very minor nit: depfile is not very suitable name for a _package does not assign semantic meaning to the contents of the map[string]map[string]string_

davecheney:
I'll probably move this package out of gb, I think it'll be called taggedkv.

re: no single format, yes I am disappointed as well.

PR #547, opened 2016-01-18
Upstream gvt restore patches: delete old deps and parallel downloads

FiloSottile:
gvt had a command called rebuild, which looked exactly like restore (because I'm an idiot and didn't remember removing restore when doing the first port).

I renamed it restore, and this PR upstreams the two main changes that were applied to gvt rebuild:

 * old dependencies are deleted if they exist before the copypath
 * downloads are parallel (FiloSottile/gvt#10 by @petrkotek)


cmd/gb-vendor/restore.go
```diff
@@ -11,23 +14,35 @@ import (
 	"github.com/constabulary/gb/vendor"
 )
 
+var (
+	rbInsecure    bool // Allow the use of insecure protocols
```
davecheney: I cannot see where this is being set

cmd/gb-vendor/restore.go
```diff
@@ -11,23 +14,35 @@ import (
...
+	rbConnections uint // Count of concurrent download connections
```
davecheney: or this

davecheney:
Look, this is going to be an awkward conversation, but I don't want to take any patches for gb vendor restore. It was a mistake to accept that feature in the first place because it confuses the story that gb handles dependencies through vendoring into `$PROJECT/vendor/src`.

If you don't want to, or cannot vendor the source into `$PROJECT/vendor/src` then I offer the depfile proposal, #536 as an alternative.

Thanks for your understanding.

cmd/gb-vendor/restore.go
FiloSottile: Because I'm terrible. Fixed.

FiloSottile:
I fully expected that reply. I'm equally unhappy about gvt restore. However, these are patches that I had accepted a long time ago in gvt, and I'm really trying to keep the core of gvt in sync with gb-vendor (for my own sanity).

I believe these are pretty technical changes, nothing behavioral, so I gave upstreaming a try. I'd still be happier if this was accepted, after which I'd stop taking patches on gvt restore, too. But I'll understand if you close it, and I'll just keep the commits around and apply them to gb master if I need it when syncing.

PR #548, opened 2016-01-18
Only match package from $PROJECT/src

### LINKED ISSUE #505: cmd/gb: refuse to build any package that is not rooted in $PROJECT/src
Because gb uses `go/build` package resolution can be fooled into building any package that is in scope. This should be prohibited.
```
zapf(/tmp/src) % gb build net/http
FATAL: command "build" failed: failed to resolve import path "net/http": cannot find package "golang.org/x/net/http2/hpack" in any of:
        /Users/dfc/go/src/golang.org/x/net/http2/hpack (from $GOROOT)
        /tmp/src/golang.org/x/net/http2/hpack (from $GOPATH)
        /tmp/vendor/src/golang.org/x/net/http2/hpack
```
Had this not been Go 1.6, this would have succeeded, and will probably succeed once the resolver refactorings land.

davecheney:
Fixes #505

Only packages matching $PROJECT/src/... will match package arguments or globs.

### DIFF
--- a/cmd/gb/gb_test.go
+++ b/cmd/gb/gb_test.go
@@ -1541,3 +1541,22 @@ func main() { println("hello world") }
 	gb.run("build")
 	gb.mustNotExist(gb.path("bin", name))
 }
+
+func TestGbBuildCannotReferencePackagesOutsideProjectSrc(t *testing.T) {
+	gb := T{T: t}
+	defer gb.cleanup()
+
+	gb.tempDir("src/vendor/src/net/http")
+	gb.tempFile("src/vendor/src/net/http/http.go", `package http
+func init() {
+	println("I should not compile")
+}
+`)
+	gb.cd(gb.tempdir)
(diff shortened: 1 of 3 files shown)


sr:
:sparkling_heart: thanks for fixing this.

sr:
Side effects, possibly unwanted, probably interesting to note regardless:

Prior to this change, I was able to vendor all "binaries", e.g. do

```
$ gb vendor fetch github.com/golang/protobuf
$ gb build github.com/golang/protobuf/protoc-gen-go
$ ls bin/protoc-gen-go
```

davecheney:
@sr sorry that this broke your workflow. I think I made  promise to someone that I would let package resolution match both roots in `$PROJECT/src` and `$PROJECT/vendor/src`, i'll fix it this morning.

davecheney:
@sr #555 should fix the problem, i've added some tests to make sure

PR #555, opened 2016-01-19
Fix package resolution (again)

### LINKED ISSUE #505: cmd/gb: refuse to build any package that is not rooted in $PROJECT/src

davecheney:
Fixes #505

Refactoring the importer and package root resolution broke a
supported, or promised (I cannot find the link) mode of operation.
See https://github.com/constabulary/gb/pull/548#issuecomment-172972528

This change restores the ability to build packages from $PROJECT/vendor/src
providing they are named explicitly, globbing is not supported.

### DIFF
--- a/cmd/gb/build.go
+++ b/cmd/gb/build.go
@@ -151,9 +151,8 @@ func resolveRootPackages(r Resolver, paths ...string) ([]*gb.Package, error) {
 		if err != nil {
 			return pkgs, fmt.Errorf("failed to resolve import path %q: %v", path, err)
 		}
-		if pkg.SrcRoot != filepath.Join(r.Projectdir(), "src") {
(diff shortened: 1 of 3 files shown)


sr:
Confirmed this fixes the issue. Thanks!

PR #581, opened 2016-03-19
gb-vendor: fix recursive dependency fetching with -branch

FiloSottile:
The branch indication was being carried on to the recursive dependencies
because the global branch variable was getting shadowed before being
cleared.

Reported in FiloSottile/gvt#25


davecheney:
Nice one. Thank you.

PR #578, opened 2016-03-12
support ssh with bitbucket and private repos

sc0rp1us:
P.S. do not forget bring back vendor imports

```
diff --git a/cmd/gb-vendor/fetch.go b/cmd/gb-vendor/fetch.go
index ee818be..711f1dd 100644
--- a/cmd/gb-vendor/fetch.go
+++ b/cmd/gb-vendor/fetch.go
@@ -13,7 +13,7 @@ import (
        "github.com/constabulary/gb"
        "github.com/constabulary/gb/cmd"
        "github.com/constabulary/gb/fileutils"
-       "github.com/constabulary/gb/vendor"
+       "github.com/sc0rp1us/gb/vendor"
 )

```

PR #586, opened 2016-04-01
Fix problem with private repositories

girokon:
If I use my private repository at github as dependency in my project, when I try to do `gb restore` or `gb update`, gb ignores my schema, and by default uses https and this breaks my build process on CI, because git asks to input a username and a password for https and I can't automatically answer on these questions in the build script.
This PR solves the issue

### DIFF
--- a/cmd/gb-vendor/restore.go
+++ b/cmd/gb-vendor/restore.go
@@ -9,6 +9,7 @@ import (
 	"github.com/constabulary/gb/cmd"
 	"github.com/constabulary/gb/fileutils"
 	"github.com/constabulary/gb/vendor"
+	"net/url"
 )
 
 func addRestoreFlags(fs *flag.FlagSet) {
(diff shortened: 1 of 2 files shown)


davecheney:
I'm not sure this will work. I don't believe the scheme is recorded in the manifest. 

Please add a test to support this change, and ensure it doesn't regress later.

davecheney:
fwiw. I think you probably want to use the .gitconfig method that is required by the go tool to force ssh on certain repos.

Or, just skip gb-vendor if it's not working for you, use git submodules instead.

davecheney:
Thanks. Sorry it took so long to merge.

davecheney:
I'm sorry, I had to revert this fix. 

Please,

1. Raise an issue describing what you see, please also look at #578 
2. Please repropose this PR including a test.

Thanks

PR #616, opened 2016-06-21
cmd/gb: restore "all" alias

davecheney:
Restore the "all" alias for matching packages. This was never advertised
but apparently people are using it[1], so hack it back in with a test.

1. https://twitter.com/archeronl/status/745202118332657664


domudall:
`gb build all` is in the Getting Started guide - https://getgb.io/examples/getting-started/ - If this is not the correct command to be using, can the guide please be updated to reflect?

davecheney:
Wow. I'm sorry, I'll update that documentation, but I've added "all" back and added a test so it doesn't regress.

Sorry about the breakage.

PR #629, opened 2016-06-29
cmd/gb: add tag key to depfile

davecheney:
Add a new depfile key, `tag` which is an unadorned github release tag.


tomasky:
But some projects  didn't   have tags

davecheney:
Then they need to start making them, and you need to push them to do so. http://dave.cheney.net/2016/06/24/gophers-please-tag-your-releases

PR #631, opened 2016-07-01
internal/vendor: Add newline to end of file

zchee:
Currently, `git-diff` command suggests "No newline at end of file".

And if edit `vendor/manifest` file uses any editor such as vim, will add newline to the end of file.
So meaningless commit diff will create.
However, this change is adding a new line to all of the manifest files of until now, so also will create meaningless commit diff.

If it normal behavior, please ignore it.

### DIFF
--- a/internal/vendor/manifest.go
+++ b/internal/vendor/manifest.go
@@ -118,6 +118,11 @@ func writeManifest(w io.Writer, m *Manifest) error {
 	if err != nil {
 		return err
 	}
+	// MarshalIndent does not append a newline after the final closing
+	// curly brace, however many editors will helpfully do this which causes
+	// lots of pointless diffs if you edit the manifest file manually. To work
+	// around this, add a newline after MarshalIndent.
+	buf = append(buf, '\n')
 	_, err = io.Copy(w, bytes.NewReader(buf))
 	return err
 }
--- a/internal/vendor/manifest_test.go
+++ b/internal/vendor/manifest_test.go
@@ -96,7 +96,8 @@ func TestEmptyPathIsNotWritten(t *testing.T) {
 			"branch": "master"
 		}
 	]
-}`
+}
+`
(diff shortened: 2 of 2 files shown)


davecheney:
The policy of this project is to raise an issue before sending a PR. In future please consider this.

davecheney:
Hmm. I see the problem, I think this is reasonable. Please fix the test failures then ping me to let me know you've updated the PR.

zchee:
> The policy of this project is to raise an issue before sending a PR. In future please consider this.

Ah, sorry...
I will be careful not to forget that in the future.

> Hmm. I see the problem, I think this is reasonable. Please fix the test failures then ping me to let me know you've updated the PR.

Thanks.
I will fix test file and ping to you later. Please wait for a little while longer.

zchee:
@davecheney Fixed `internal/vendor/manifest_test.go`.
All tests have passed. PTAL.

Thanks.

internal/vendor/manifest.go
```diff
@@ -118,6 +118,7 @@ func writeManifest(w io.Writer, m *Manifest) error {
 	if err != nil {
 		return err
 	}
+	buf = append(buf, byte('\n'))
```
davecheney: // MarshalIndent does not append a newline after the final closing
// curly brace, however many editors will helpfully do this which causes
// lots of pointless diffs if you edit the manifest file manually. To work
// around this, add a newline after MarshalIndent.
buf = append(buf, '\n')

zchee: @davecheney Got it. Fix and Added comment now.

davecheney:
Thank you. This will go out in 0.4.4 sometime in the next week.

zchee:
@davecheney Thanks for the merge!
and `gb` is very useful, thanks again for great tool.

Issue #212, opened 2015-06-11
Can we please have a version in this tool?


wolfeidau:
Currently there is no version information provided so it is difficult to assert whether the tool needs updating.

As in `gb version` just helps me keep things in sync across machines.

Thanks

davecheney:
Yup, that is a fair request.

At the moment we're moving a orbital speeds, hopefully we'll slow down to a pace where milestones and releases start to make good sense.

What will most likely happen is we'll move from go get to self hosted, and at that point ship actual releases via github.

I can't promise to do this in the short term, but I understand the requirement and think it valid.

leifmadsen:
@davecheney Just cruising by. How much more is required to close this out now that you're tagging releases like 0.1.0 and 0.1.1 on GitHub?

davecheney:
Unfortunately a lot as the go tool does not have the ability to consume a
git tag and annotate a binary. This can be done with extra scripting, but
that won't help people that just use go get.

mpasternacki:
How about scripting the release process? Let's say, a `release.sh` script, which would take version number as argument, and do something along the lines of:

```sh
#!/bin/sh
set -e
VERSION="$1"

cat >version.go <<EOF
package gb

const Version = "$VERSION"
EOF

git commit -m "Release version $VERSION" version.go

# If relevant, sign the tag with gpg
git tag -a -m "Tag version $VERSION" v$VERSION

cat >version.go <<EOF
package gb

const Version = "$VERSION+git"
EOF

git commit -m "Set development version $VERSION+git" version.go

git push --tags
```

This would take care of keeping tags and committed version number in sync. This would mean that building directly from a tagged release (e.g. by a packaging tool) has exact version number defined, and using `go get` directly off master is marked `+git`, so that work-in-progress versions are clearly distinguished. The downside is always having `+git` on `go get`-derived libraries, because master would always be younger than the tag.

A more stable/reliable option could be adopting something like git-flow, where master branch is always latest released version, and a separate develop branch is release-ready development. Then master would always be a tagged revision, develop would add `+git` suffix, and release process would drop it. This is more complex and not everybody likes git-flow, but it has the benefit of master (pulled by default by `go get`) always being the most recent released version.

mpasternacki:
(Hm. Unless `go get` pulls the default branch rather than specifically master. With git-flow-like setup you need to set _develop_ as main branch on github, otherwise all the PRs you'll get will be against master)

davecheney:
[...]

PR #639, opened 2016-07-22
Add 'gb version' tool

kevinburke:
Run `gb version` to print the version of GB, and the Go version used to compile
the gb binary. This information is useful for debugging and error reporting.

I'm not sure how you release GB - I assume a git tag is involved somewhere
- but that process will need to update to bump the version constant in
`cmd/gb/version.go`. One such tool exists here: github.com/Shyp/bump_version.

### DIFF
--- a/cmd/gb/version.go
+++ b/cmd/gb/version.go
@@ -1,0 +1,29 @@
+package main
+
+import (
+	"fmt"
+	"runtime"
+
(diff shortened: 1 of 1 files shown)


davecheney:
Thank you for this proposal. Sadly I cannot accept it at this time, please see #212 for the rational for this decision.

PR #641, opened 2016-08-17
Restore in parallel

gsquire:
I know this doesn't have a specific issue, but I figured it might be worth submitting a PR for. I found when building my project in CI, running `gb vendor restore` is the longest part.

Let me know if there are any glaring issues.


gsquire:
Hi @davecheney 

I am not sure why the tests fail for some of the versions and not others. If you have any time/insight I would appreciate it.

PR #652, opened 2016-11-03
Run compiler from the project directory

ssmccoy:
In an attempt to solve issue #406, run the compiler instead of from
the package directory from the root of the project. Make all paths
passed to the compiler relative to the project root.

This appears to work, and the tests still pass; but it corrects the
output of compiler errors and they now appear with filenames relative
to the `GB_PROJECT_DIR`.


ssmccoy:
The failure appears to be unrelated to my change.

ssmccoy:
I'd love some feedback on this.

davecheney:
Hello, 

Thanks

Dave

davecheney:
Sorry I haven't responded to this yet. Given the history of trying to fix
this in the past I'll need to find some time for a detailed review.

ssmccoy:
@davecheney I am relatively certain this isn't solving for all use-cases, but we have been using in our project (under my fork of this project) since this PR was opened and, within the usecase of our project, it solves the problem for go sources without side-effect.

I think that a partial solution is probably better than no solution, since the behaviour of gb WRT compiler errors is a huge issue for adoption. We have the need to start adding assembler to our project build, and I will ensure that similar support is added for the assembler at that time.

PR #656, opened 2016-11-10
Error when multiple paths are passed with --all

### LINKED ISSUE #646: usability: error if extra arguments passed to --all
I wanted to delete two dependencies. I tried `gb vendor delete a b` and got:

```
FATAL: command "delete" failed: delete: import path or --all flag is missing
```

I then tried: 

```
gb vendor delete --all a b
```

I expected that this would delete both a and b, but instead deleted all packages.

I think `--all` should error if any positional arguments are passed, and/or delete could have a clearer error message if multiple positional arguments are passed.

nubunto:
Previously, `gb vendor --all a b` would just delete all vendored
packages.

Closes #646


davecheney:
Hello, 

Thanks

Dave

PR #662, opened 2016-11-20
Cleanup go vet warnings

gliptak:
https://goreportcard.com/report/github.com/constabulary/gb

### DIFF
--- a/cmd/gb/gb_test.go
+++ b/cmd/gb/gb_test.go
@@ -451,7 +451,7 @@ func (t *T) mustBeEmpty(root string) {
 		found = append(found, path)
 		return nil
 	}
-	err = filepath.Walk(root, fn)
+	filepath.Walk(root, fn)
 	if len(found) > 0 {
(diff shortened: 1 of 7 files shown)


cmd/gb/gb_test.go
```diff
@@ -451,7 +451,7 @@ func (t *T) mustBeEmpty(root string) {
 		found = append(found, path)
 		return nil
 	}
-	err = filepath.Walk(root, fn)
+	_ = filepath.Walk(root, fn)
```
davecheney: this is a bug, err is not being checked.

gliptak: https://github.com/constabulary/gb/blob/master/cmd/gb/gb_test.go#L449 would have failed the test already

davecheney: what about 

filepath.Walk(root, fn)

without assigning the unused value to anything?

gliptak: changed

PR #671, opened 2016-12-02
FIX: limit goroutines in restore command

adone:
Our project have ~120 dependencies and gb creates ~120 parallel requests to github.com.
As result: `curl: (56) SSLRead() return error -36` and `EOF`


davecheney:
Sorry for the delay.

This is a lot of code, can it be made simpler by adding a simple semaphore using a buffered channel?

davecheney:
Obsoleted by #695

Issue #690, opened 2017-02-17
go1.8 and gb test with latest release (v0.4.3) do not play nice.


freeformz:
It fails with something like:

Given this [fixture from the heroku buildpack](https://github.com/heroku/heroku-buildpack-go/tree/master/test/fixtures/gb-with-tests/src) I get this when using gb @ v0.4.3:

```console
$ gb test
# cmd/fixture/testmain
_testmain.go:48: cannot use matchString (type func(string, string) (bool, error)) as type testing.testDeps in argument to testing.MainStart:
	func(string, string) (bool, error) does not implement testing.testDeps (missing MatchString method)
# sub/testmain
_testmain.go:56: cannot use matchString (type func(string, string) (bool, error)) as type testing.testDeps in argument to testing.MainStart:
	func(string, string) (bool, error) does not implement testing.testDeps (missing MatchString method)
FATAL: command "test" failed: exit status 2
```

This does not happen when I checkout master, recompile gb and re-run the tests.

Can we get a new release?

davecheney:
I think this is fixed with #692. Please reopen if this is not the case. A test case was added to the integration test suite based on the sample code in the error report.

freeformz:
Thanks, can I get a new release ? That was my main concern, to have a released version that supports go1.8. Thanks for all of your work!

PR #691, opened 2017-02-21
.travis.yml: add 1.8.x builds to travis

davecheney:


davecheney:
Test flake caused by gopkg.in. Updates #690

PR #692, opened 2017-02-21
remove old go1.8 NoGoError code

davecheney:
Removes the since rolled back change to build.NoGoError Go 1.8+ workaround.

Fixes gb#690


davecheney:
@freeformz i'm going to commit this to fix #690. I grabbed the test fixture from the bug report and it passes fine on all versions of Go from 1.4 to tip.

PR #695, opened 2017-03-01
Add semaphore to `gb vendor restore`

### LINKED ISSUE #698: `gb vendor restore` fails for a large set of dependencies
Using `go vendor restore` on a project with a large number of dependencies (in my case, 100+) never succeeds. It seems like some resource ran out and random errors happened, like:

```
Getting google.golang.org/genproto/googleapis/api/label
Getting gopkg.in/tomb.v1
Getting gopkg.in/yaml.v2
skipping insecure protocol: git://github.com/spf13/afero
skipping insecure protocol: http://github.com/spf13/afero
fatal: unable to access 'https://code.googlesource.com/google-api-go-client/': Failed to connect to code.googlesource.com port 443: Operation timed out
```

**Version:** HEAD
**Platform:** macOS Sierra (gb compiled with Go 1.8)

I've proposed a fix for this in #695, which limits parallelization to the number of CPUs. It's still quite fast and has been incredibly reliable for me.

andremedeiros:
For projects with a lot of dependencies, `gb vendor restore` won't work all the time and fail silently.

This makes pulling them more manageable and more resilient.

Closes #698

### DIFF
--- a/cmd/gb-vendor/restore.go
+++ b/cmd/gb-vendor/restore.go
@@ -13,26 +13,36 @@ import (
 	"github.com/pkg/errors"
 )
 
+const DefaultJobs = 8
+
 func addRestoreFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&insecure, "precaire", false, "allow the use of insecure protocols")
+	fs.IntVar(&jobs, "jobs", DefaultJobs, "maximum amount of restoration jobs occurring at the same time")
 }
 
-var cmdRestore = &cmd.Command{
-	Name:      "restore",
-	UsageLine: "restore [-precaire]",
-	Short:     "restore dependencies from the manifest",
-	Long: `Restore vendor dependencies.
+var (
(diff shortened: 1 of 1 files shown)


dcheney-atlassian:
Thanks for fixing this. This makes me kind of mad actually. gb vendor restore used to work perfectly, slowly, reliably, perfectly. Then we had to go and make it faster, and now it is less reliable. I am unconvinced that this was an improvement.

I'm tempted to revert the original parallisation change.

dcheney-atlassian:
Oh, and you _must_ raise a bug before proposing a PR. Thank you.

andremedeiros:
> I'm tempted to revert the original parallisation change.

This is still good to have. The problem I was facing was that I worked on a project with over 100 dependencies and it gets hairy for those cases.

I'll also raise a bug.

jeromer:
Do you plan to include this patch ? It would be super useful.

PR #696, opened 2017-03-07
+ `-o` flag to set output binary file path for `gb build`. fixes #309

### LINKED ISSUE #309: gb build a binary with given name
Similar to "go build" supports -o flag, I feel it would be better if "gb build" supports specifying binary name:

gb build -o myserver package/.../crm_server

Thanks.

mkorolyov:
When we pass tags through gb build for project foo (e.g. `gb build -tags 'production monitoring'`) gb generates bin file `foo-production-monitoring` while we expects just `foo`.

The idea is to add `-o` flag to specify output file name(e.g. `db build -tags 'production monitoring' -o SOME_PATH/foo'`)

### DIFF
--- a/cmd/gb/build.go
+++ b/cmd/gb/build.go
@@ -45,7 +45,7 @@ var (
 
 func addBuildFlags(fs *flag.FlagSet) {
 	// TODO(dfc) this should accept a *gb.Context
-	fs.StringVar(&outputFileTemplate, "o", "", "output file template. possible placeholders: target_path, tags: example: `gb build -o '{{.target_path}}-{{.tags}}'`")
+	fs.StringVar(&outputFileTemplate, "o", "", "output file template")
 	fs.BoolVar(&R, "r", false, "perform a release build")
 	fs.BoolVar(&F, "f", false, "rebuild up-to-date packages")
 	fs.BoolVar(&FF, "F", false, "do not cache built packages")
@@ -67,6 +67,10 @@ dependencies.
 
 Flags:
 
+	-o
+		changes output binary file name. golang template expected.
+		possible placeholders: target_path, tags.
+		example: gb build -o '{{.target_path}}-{{.tags}}'
 	-f
 		ignore cached packages if present, new packages built will overwrite
 		any cached packages. This effectively disables incremental
--- a/cmd/gb/build.go
+++ b/cmd/gb/build.go
@@ -39,10 +39,13 @@ var (
 	dotfile string // path to dot output file
 
 	buildtags []string
+
+	outputFileName string
 )
 
 func addBuildFlags(fs *flag.FlagSet) {
 	// TODO(dfc) this should accept a *gb.Context
+	fs.StringVar(&outputFileName, "o", "", "output file name")
 	fs.BoolVar(&R, "r", false, "perform a release build")
 	fs.BoolVar(&F, "f", false, "rebuild up-to-date packages")
 	fs.BoolVar(&FF, "F", false, "do not cache built packages")
--- a/cmd/gb/build.go
+++ b/cmd/gb/build.go
@@ -40,12 +40,12 @@ var (
 
 	buildtags []string
 
-	outputFileName string
+	outputFileTemplate string
 )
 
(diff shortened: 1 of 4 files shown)


mkorolyov:
any comments/thoughts/etc?

davecheney:
I'm not keen on adding this, gb works on _projects_ not _packages_, -o implies a specialisation on a single output binary which I don't feel comfortable with.

mkorolyov:
@davecheney Hi Deve,
Didn't get why you mentioned packages here.
Problem that i wan't to solve is when we build i project with set of flags i don't want the output binary name to change with `-flag1-flag2-flag3-etc`.
Simple option `-o` implemented for a wide list of tools that produce something, why gb should be exception?

mkorolyov:
@davecheney  
What if we will ad possibility to set format of output for binary files? like `${project_name}-${tags}` or `${project_name}_{$arch_name}`?

davecheney:
gb works on projects, and projects can have multiple outputs. Adding -o would put gb in the same position as the go tool; go build -o only works in some cases.

davecheney:
Doesn't that exist already? 

What is the problem you are trying to solve?

mkorolyov:
> What is the problem you are trying to solve?
when we use build flags we got binaries with formatting `projectname-flag1-flag2` while need just simple `projectname`

currently there is a possibility only to set output bin dir.

davecheney:
> when we use build flags we got binaries with formatting projectname-flag1-flag2 while need just simple projectname

You get `command-flag`, not `project-name`, flag. If there is only one command in your project and it is the name of the project, this will be the same thing.

> currently there is a possibility only to set output bin dir.

The output directory is hard coded, that is why files with a different tag have a different suffix.

What would you say to 

$PROJECT/bin/tag/tag/tag/binary

so

`gb build -tag release` would generate `$PROJECT/bin/release/command`

mkorolyov:
@davecheney what if we have more than 1 build tag? 5 for example? it is not convenient to build by hand binary path that depends on specified go build flags.

davecheney:
> @davecheney what if we have more than 1 build tag? 5 for example? it is not convenient to build by hand binary path that depends on specified go build flags.

But you would know the path, it's the same as the flags you passed in.

Please, there is something you are not telling me, something about how your Ci or build and release process works that has put you in a position of needing to change how gb works. If you can explain what the problem is that you are solving, maybe there is a different way to achieve what you need.

mkorolyov:
@davecheney 
> gb build -tag release would generate $PROJECT/bin/release/command

didn't find such logic in the code, could you pls show it? Asking because current gb produced `$project/bin/projectname-release` for me and not `$project/bin/release/projectname`

davecheney:
Yes, sorry for the confusion, this was an alternative proposal. It is not implemented.

mkorolyov:
@davecheney  so how about proposal for possibility to set output binary format?

davecheney:
I'm sorry, I don't want to add this flag for the reasons I've explained above. 

We can talk about alternatives.

mkorolyov:
@davecheney 
i am already proposing an alternative
i mean add something like --output-bin-format with usage
`gb build --output-bin-format '${name}-${tags}' -tags release` 
or `gb build --output-bin-format '${name}-${arch}'` 
or `gb build --output-bin-format '${name}'`
which produce 
`$projectdir/bin/projectname-release`
or $projectdir/bin/projectname-arm`
or `$projectdir/bin/projectname`

davecheney:
No, I'm sorry I won't want to add that. It's a more complicated version of -o

mkorolyov:
@davecheney it is adoption of -o for gb multi binary semantics.

What would you propose than to achieve same result?

davecheney:
My counter proposal is

1. do nothing, the name of the final binary is predictable from the inputs
to gb build
2. -or-, mvoe from $PROJECT/bin/command-tag-tag-tag to
$PROJECT/bin/tag/tag/tag/command

davecheney:
Thank you for this pr, but I am not going to merge. It. As I explained I do not want to make -o configurable. I am going to close this pr and the issue.


Thank you for you understanding

Issue #709, opened 2017-05-02
Cross complation from amd64 -> 386 is broken at tip


davecheney:
gb cannot handle cross compilation between intel architectures at tip (go 1.9) Actually it's broken on all cross compilation targets, but amd64/386 are one of the few that can be cross compiled, then run locally.

- [ ] add tests for cross compilation
- [ ] fix test breakage on tip / go 1.9

PR #713, opened 2017-05-02
gb: Annotate internal/poll as having external files

davecheney:
Go 1.9 adds a few more internal packages that are implemented with
//go:linkname.


davecheney:
Update #709

PR #716, opened 2017-05-08
Add CXXFiles to TestPackage

lukejohnston:
With a project layout like this:
```
.
└── src
    └── proj
        ├── a
        │   ├── a.cpp
        │   ├── a.go
        │   └── a.h
        └── main.go
```
the `test` command will give a linker error when processing the `a` directory. This is because the `.cpp` files are not added to the test package. This change adds the `.cpp` files to the test package.


davecheney:
Thank you. Can you please add a test to cmd/gb/gb_test.go

lukejohnston:
Sure, done. Thanks for the quick reply.

davecheney:
LGTM. Thanks.

PR #726, opened 2017-06-29
Treat CGO_ENABLED="0" the same as cross-compiling.

NegativeMjark:
This means that explicitly disabling cgo using CGO_ENABLED="0" in
order to get a static binary will work even when the libraries in
GOROOT have been compiled with cgo enabled.

This could be an easy work around for #328


MrSaints:
Any updates on this?

PR #730, opened 2018-01-10
vendor: respect repository field in restore and update

cryptix:
Somehow the `repository` field from the manifest isn't used.

With this I can replace packages with forks.

### DIFF
--- a/cmd/gb-vendor/restore.go
+++ b/cmd/gb-vendor/restore.go
@@ -74,8 +74,8 @@ func restore(ctx *gb.Context) error {
 func restoreWorker(ctx *gb.Context, work chan vendor.Dependency, wg *sync.WaitGroup, errChan chan error) {
 	defer wg.Done()
 	for dep := range work {
-		fmt.Printf("Getting %s\n", dep.Importpath)
-		repo, _, err := vendor.DeduceRemoteRepo(dep.Importpath, insecure)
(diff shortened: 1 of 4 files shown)


davecheney:
I deeply regret adding restore to gb vendor, thanks for fixing this, I’ll take a look soon.

cryptix:
> I deeply regret adding restore to gb vendor, ...

oh why is that?

I might not be best style but I tend to keep small patches inside my tree until I move them to PRs/external forks. It reduces the overhead of dealing with a lot more repos for me and the restore command is pretty helpful to get back the upstream state.

davecheney:
It has been the biggest source of bugs, especially when people became so reliant on it that parallel downloading of artifacts had to be implemented.

cryptix:
~~I just noticed an issue with my approach. Consider these two entries in `manifest`, they retrieve two packages from the same repository.~~

~~Since I just replaced `importPath` with `repository`, it now clones the root of the repositories into the two `improtpath` locations, leading to a broken vendor state.~~

Disregard the previous comment. Somehow my `vendor/manifest` didn't have `"path": "..."` entries to build the correct copy.

cryptix:
Testing this some more, I noticed that I broke vanity imports.

I pushed a really dirty hack in 0153a33 to make `golang.org/x/...` work again but it's not good since it only works for `go.googlesource.com/...`.

